Issue #486, opened 2017-03-29
Refactoring project structure (3.3.0) 


reckart:
Further refactoring project structure for modularity.

PR #510, opened 2017-06-02

astier:
- Use PropertyModel instead of raw data

### DIFF
--- a/webanno-ui-annotation/src/main/java/de/tudarmstadt/ukp/clarin/webanno/ui/annotation/detail/AnnotationFeatureForm.java
+++ b/webanno-ui-annotation/src/main/java/de/tudarmstadt/ukp/clarin/webanno/ui/annotation/detail/AnnotationFeatureForm.java
@@ -241,7 +241,6 @@ public class AnnotationFeatureForm
                 aTarget.add(featureEditorPanelContent.get(0));
             }
         });
-        textfield.setOutputMarkupId(true);
         textfield.add(new AttributeAppender("style", "opacity:0", ";"));
         // forwardAnno.add(new AttributeAppender("style", "filter:alpha(opacity=0)", ";"));
         return textfield;
@@ -281,7 +280,8 @@ public class AnnotationFeatureForm
 
     private LayerSelector createDefaultAnnotationLayerSelector()
     {
-        return new LayerSelector("defaultAnnotationLayer", annotationLayers);
+        return new LayerSelector("defaultAnnotationLayer",
+            new PropertyModel<>(this, "annotationLayers"));
     }
 
     private Label createSelectedTextLabel()
@@ -592,11 +592,10 @@ public class AnnotationFeatureForm
     {
         private static final long serialVersionUID = 2233133653137312264L;
 
-        LayerSelector(String aId, List<? extends AnnotationLayer> aChoices)
+        LayerSelector(String aId, IModel<List<? extends AnnotationLayer>> aChoices)
         {
-            super(aId, aChoices);
+            super(aId, aChoices, new ChoiceRenderer<>("uiName"));
             setOutputMarkupId(true);
-            setChoiceRenderer(new ChoiceRenderer<>("uiName"));
             add(new AjaxFormComponentUpdatingBehavior("change")
             {
                 private static final long serialVersionUID = 5179816588460867471L;


webanno-ui-annotation/src/main/java/de/tudarmstadt/ukp/clarin/webanno/ui/annotation/detail/AnnotationFeatureForm.java
```diff
@@ -275,8 +248,7 @@ public void onSubmit(AjaxRequestTarget aTarget, Form<?> aForm)
             }
         });
 
-        add(layerSelector = new AnnotationFeatureForm.LayerSelector
-            ("defaultAnnotationLayer", annotationLayers));
+        add(layerSelector = new LayerSelector(this, "defaultAnnotationLayer", annotationLayers));
```
reckart: Instead of directly passing the `annotationLayers` as a list, an IModel instance should be passed.

webanno-ui-annotation/src/main/java/de/tudarmstadt/ukp/clarin/webanno/ui/annotation/detail/LayerSelector.java
```diff
@@ -0,0 +1,106 @@
...
+        setOutputMarkupId(true);
+        setChoiceRenderer(new ChoiceRenderer<>("uiName"));
+        add(new AjaxFormComponentUpdatingBehavior("change")
+        {
+            private static final long serialVersionUID = 5179816588460867471L;
+
+            @Override
+            protected void onUpdate(AjaxRequestTarget aTarget)
+            {
+                AnnotationDetailEditorPanel editorPanel = featureForm.getEditorPanel();
```
reckart: Instead of passing the AnnotationDetailEditorPanel, passing an AnnotationActionHandler (which is implemented by AnnotationDetailEditorPanel) would seem more appropriate. Also passing an IModel<AnnotatorState> separately instead of fetching it from the ADEP.

webanno-ui-annotation/src/main/java/de/tudarmstadt/ukp/clarin/webanno/ui/annotation/detail/AnnotationFeatureForm.java
astier: Can we make a separate issue for this? I am not familiar with IModel. It also wasn't me who implemented this. Maybe we can fix this an other time.

webanno-ui-annotation/src/main/java/de/tudarmstadt/ukp/clarin/webanno/ui/annotation/detail/LayerSelector.java
astier: It seems like the ADEP is needed in this method, because an ADEP specific method is called at line 103:
```
editorPanel.clearFeatureEditorModels(aTarget);
```
 Also, the ADEP is passed as an argument at line 80 to the DeleteOrReplaceAnnotationModalPanel instance what has to be also of type ADEP because the DeleteOrReplaceAnnotationModalPanel instance calls ADEP specific methods.

reckart: True. I'll see how far I can clean up/push this when I merge.

webanno-ui-annotation/src/main/java/de/tudarmstadt/ukp/clarin/webanno/ui/annotation/detail/AnnotationFeatureForm.java
```diff
@@ -282,7 +282,33 @@ protected void onConfigure()
...
+
+            @Override
+            public void setObject(List<? extends AnnotationLayer> annotationLayers)
+            {
+                this.annotationLayers = annotationLayers;
+            }
+
+            @Override
+            public void detach()
+            {
```
astier: Not sure what I should do here.

astier:
I dropped my original changes where I refactored LayerSelector.class in its own file. Instead I tried to follow your request and to make use of an IModel instead of passing the raw data to the DropDownChoice instance. I hope I've done it right.

webanno-ui-annotation/src/main/java/de/tudarmstadt/ukp/clarin/webanno/ui/annotation/detail/AnnotationFeatureForm.java
reckart: Use something like PropertyModel.of(this, "annotationLayers")?

astier:
I changed IModel to PropertyModel like you suggested.

astier:
This one is ready.

PR #517, opened 2017-06-07
layer behaviour options

remstef:
- added two options
1. allow only zero width span options
2. run javascript action on click

- run defined on click events in brat annotation editor


reckart:
Could you please open issue for the new features and then submit one PR per issue? The commit message should have the format

```
#123 - Title of the issue
- some change detail
- other chain detail
```

remstef:
Sure, here we go.
Related to issues #521 and #520

PR #525, opened 2017-06-12

astier:
- Replace Stringbuffer with StringBuilder
- Remove unnecessary unboxing
- Remove unnecessary for-loop

### DIFF
--- a/webanno-api-dao/src/main/java/de/tudarmstadt/ukp/clarin/webanno/api/dao/AnnotationSchemaServiceImpl.java
+++ b/webanno-api-dao/src/main/java/de/tudarmstadt/ukp/clarin/webanno/api/dao/AnnotationSchemaServiceImpl.java
@@ -799,14 +799,9 @@ public class AnnotationSchemaServiceImpl
     @Transactional
     public List<Tag> listTags(TagSet aTagSet)
     {
-        List<Tag> tags = entityManager
+        return entityManager
                 .createQuery("FROM Tag WHERE tagSet = :tagSet ORDER BY name ASC", Tag.class)
                 .setParameter("tagSet", aTagSet).getResultList();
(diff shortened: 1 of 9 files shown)


webanno-api-dao/src/main/java/de/tudarmstadt/ukp/clarin/webanno/api/dao/AnnotationSchemaServiceImpl.java
```diff
@@ -803,8 +803,8 @@ private AnnotationFeature createFeature(String aName, String aUiname, Project aP
                 .createQuery("FROM Tag WHERE tagSet = :tagSet ORDER BY name ASC", Tag.class)
                 .setParameter("tagSet", aTagSet).getResultList();
         // FIXME ?!? This loop appears to make absolutely not sense!
```
astier: I think this loop can actually be removed.

astier: Whoops. The loop below which you can not see.

reckart: Probably yes...

astier:
I changed now for this PR the commit message by squashing all the commits. I also removed the mentioned unnecessary for-loop.

astier:
Since I replaced StringBuffer with StringBuilder can it cause any problems because StringBuilder is not Synchronized?

reckart:
> Since I replaced StringBuffer with StringBuilder can it cause any problems because StringBuilder is not Synchronized?

I don't think so.

Issue #384, opened 2016-10-14
Automatic sign off from WebAnno after session expiration


GiantEnemyCrab:
When the session expires, WebAnno should automatically sign off users and send it to home screen.

I am experiencing that WebAnno stays open (even after session has expired on the server side), and if I take any action, then it finally signs me off, sending to home screen.

This is more for security reasons. There may be documents that contain sensitive data, and as an extra  cautious approach, WebAnno should automatically redirect to the home page. This is similar to these bank websites that send users to home login screen after checking account page stays open for a while without any action being done on the website.

I am not sure if "session" is technically the correct terminology in this context. But it is the logic that handles sign-off.

reckart:
I have added a countdown on the browser-side that triggers a page reload (a few seconds) after the configured server-side session timeout has expired. It only reloads the current page, but that effectively redirects to the login page.

GiantEnemyCrab:
Thank you for the change, confirmed that it is working

reckart:
@TFeige reports in #380 that auto-logout has problems in the standalone version: the counter is always at 0:00. Need to check this.

reckart:
Looks like in Winstone, the default session timeout is 0. In this case, just don't display the timer.

PR #526, opened 2017-06-12

astier:
- Replace explicit array type with <>
- Collapse redundant catch blocks
- Inline redundant variable
- Correct misspelling
- Remove unnecessary import
- Make use of Comparator.comparing()
- Remove unnecessary String.format()


reckart:
Commit messages don't match convention.

reckart:
Why do you link these commits to issue #384 *Whole project export does not include in progress curation* ?!

reckart:
Commit messages should refer issue #486 - Refactoring project structure (3.3.0) like

```
- Java 7 features
- Replace explicit array type with <>
- Collapse redundant catch blocks
```

astier:
Commit message fixed.

PR #528, opened 2017-06-12

astier:
- Replace deprecated method ObjectUtils.equals with Objects.equals
- Replace if-else blocks with switch
- Fix some comments
- Simplify code
- Add an empty line to the end of the file
- Remove explicit types with <>
- Remove unnecessary casting


reckart:
Please stick to the usual commit message format

```
#<issue number> - <issue title>
- <change 1>
- <change 2>
- ...
```

webanno-api-annotation/src/main/java/de/tudarmstadt/ukp/clarin/webanno/api/annotation/util/WebAnnoCasUtil.java
```diff
@@ -22,12 +22,7 @@
 import static org.apache.uima.fit.util.JCasUtil.selectCovered;
 import static org.apache.uima.fit.util.JCasUtil.selectFollowing;
 
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.HashSet;
-import java.util.Iterator;
-import java.util.List;
-import java.util.Set;
+import java.util.*;
```
reckart: No star imports - always specific imports.

webanno-api-annotation/src/main/java/de/tudarmstadt/ukp/clarin/webanno/api/annotation/util/WebAnnoCasUtil.java
```diff
@@ -71,15 +66,7 @@
...
-        if (a == null || b == null) {
-            return false;
-        }
-
-        if (a.getCAS() != b.getCAS()) {
-            return false;
-        }
-
-        return getAddr(a) == getAddr(b);
+        return !(a == null || b == null) && a.getCAS() == b.getCAS() && getAddr(a) == getAddr(b);
```
reckart: The previous code was more verbose but easier to read.

webanno-api-annotation/src/main/java/de/tudarmstadt/ukp/clarin/webanno/api/annotation/model/VID.java
```diff
@@ -281,12 +281,6 @@ public boolean equals(Object obj)
...
             return false;
         }
-        if (slot != other.slot) {
-            return false;
-        }
-        if (subAnnotationId != other.subAnnotationId) {
-            return false;
-        }
-        return true;
+        return slot == other.slot && subAnnotationId == other.subAnnotationId;
```
reckart: The previous code was more verbose but easier to read.

webanno-api-annotation/src/main/java/de/tudarmstadt/ukp/clarin/webanno/api/annotation/util/WebAnnoCasUtil.java
```diff
@@ -857,7 +838,7 @@ public static void setFeature(FeatureStructure aFS, AnnotationFeature aFeature,
                 aFS.setStringValue(feature, (String) aValue);
                 break;
             case CAS.TYPE_NAME_BOOLEAN:
-                aFS.setBooleanValue(feature, aValue != null ? (boolean) aValue : false);
+                aFS.setBooleanValue(feature, aValue != null && (boolean) aValue);
```
reckart: The effect is the same, but more difficult to understand than the previous code which basically read as intended: "if the value is !null, then use the value, otherwise set to false". What you have now reads as "if value is !null and value" which is much more difficult to understand.

webanno-io-tsv/src/main/java/de/tudarmstadt/ukp/clarin/webanno/tsv/util/AnnotationUnit.java
```diff
@@ -49,20 +49,18 @@ public int hashCode() {
...
 
 	@Override
-	public boolean equals(Object obj) {
-		if (this == obj)
+	public boolean equals(Object obj)
+	{
+		if (this == obj) {
 			return true;
-		if (obj == null)
+		}
```
reckart: The curly braces are part of the code style. The should stay. In cases they are missing, they should be added.

astier: They seem to be ok. IntelliJ probably auto-corrected their placement because they didn't fit to the code style.

PR #534, opened 2017-06-20

astier:
- Remove unnecessary unboxing
- Collapse redundant catch blocks
- Simplify comperator using Comperator.comparing
- Replace loop with Collection.removeIf()
- Replace anonymous type with lambda
- Replace lambda with method reference
- Remove redundant String.toString()
- Remove redundant String.format()
- Replace StringBuffer with String
- Use bulk operation instead of iteration
- Remove unnecessary array creation
- Remove redundant casts
- Remove explicit type argument
- Remove unnecessary import
- Remove unnecessary semicolon
- Format lines for better readability
- Inline unnecessary variable

### DIFF
--- a/webanno-brat/src/main/java/de/tudarmstadt/ukp/clarin/webanno/brat/annotation/BratAnnotationEditor.java
+++ b/webanno-brat/src/main/java/de/tudarmstadt/ukp/clarin/webanno/brat/annotation/BratAnnotationEditor.java
@@ -410,33 +410,32 @@ public class BratAnnotationEditor
         aResponse.render(JavaScriptHeaderItem.forReference(BratVisualizerUiResourceReference.get()));
         aResponse.render(JavaScriptHeaderItem.forReference(BratAnnotatorUiResourceReference.get()));
         //aResponse.render(JavaScriptHeaderItem.forReference(BratUrlMonitorResourceReference.get()));
-        
-        StringBuilder script = new StringBuilder();
-        // REC 2014-10-18 - For a reason that I do not understand, the dispatcher cannot be a local
+
+	// REC 2014-10-18 - For a reason that I do not understand, the dispatcher cannot be a local
         // variable. If I put a "var" here, then communication fails with messages such as
         // "action 'openSpanDialog' returned result of action 'loadConf'" in the browsers's JS
         // console.
-        script.append("(function() {");
-        script.append("var dispatcher = new Dispatcher();");
-        // Each visualizer talks to its own Wicket component instance
-        script.append("dispatcher.ajaxUrl = '" + requestHandler.getCallbackUrl() + "'; ");
-        // We attach the JSON send back from the server to this HTML element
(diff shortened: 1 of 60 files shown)


webanno-brat/src/main/java/de/tudarmstadt/ukp/clarin/webanno/brat/annotation/BratAnnotationEditor.java
```diff
@@ -409,33 +409,32 @@ public void renderHead(IHeaderResponse aResponse)
         aResponse.render(JavaScriptHeaderItem.forReference(BratVisualizerUiResourceReference.get()));
         aResponse.render(JavaScriptHeaderItem.forReference(BratAnnotatorUiResourceReference.get()));
         //aResponse.render(JavaScriptHeaderItem.forReference(BratUrlMonitorResourceReference.get()));
-        
-        StringBuilder script = new StringBuilder();
+
```
reckart: Can you please undo the changes to the string building here? The comments are no longer where they belong.

webanno-constraints/src/main/java/de/tudarmstadt/ukp/clarin/webanno/constraints/ConstraintsServiceImpl.java
```diff
@@ -217,17 +217,15 @@ public ParsedConstraints loadConstraints(Project aProject)
...
-                        errorMessage.append(e.getKey());
-                        errorMessage.append("\", conflicting values are \"");
-                        errorMessage.append(e.getValue());
-                        errorMessage.append("\" & \"");
-                        errorMessage.append(merged.getImports().get(e.getKey()));
-                        errorMessage.append(
-                                "\". Please contact Project Admin for correcting this. Constraints feature may not work.");
-                        errorMessage.append("\nAborting Constraint rules merge!");
-                        throw new ParseException(errorMessage.toString());
+                        String errorMessage = "Conflict detected in imports for key \"" +
```
reckart: I am not sure the changes here improve the situation. It would probably be best to change this into a String.format(...) call.

webanno-constraints/src/main/java/de/tudarmstadt/ukp/clarin/webanno/constraints/model/FSFPair.java
```diff
@@ -86,12 +86,11 @@ else if (!featureStructure.equals(other.featureStructure)) {
...
     public String toString()
     {
-        StringBuilder builder = new StringBuilder();
-        builder.append("FSFPair [featureStructure=");
-        builder.append(featureStructure);
-        builder.append(", affectedFeature=");
-        builder.append(affectedFeature);
-        builder.append("]");
-        return builder.toString();
+        String builder = "FSFPair [featureStructure=" +
```
reckart: I am not sure the changes here improve the situation. It would probably be best to change this into a String.format(...) call... or at least to format the code to remove the many line breaks.

PR #537, opened 2017-06-22

astier:
- Make use of StringBuilder.append() instead concatenation
- Inline variable and remove cast

### DIFF
--- a/webanno-io-tsv/src/main/java/de/tudarmstadt/ukp/clarin/webanno/tsv/WebannoTsv1Reader.java
+++ b/webanno-io-tsv/src/main/java/de/tudarmstadt/ukp/clarin/webanno/tsv/WebannoTsv1Reader.java
@@ -243,7 +243,7 @@ public class WebannoTsv1Reader
         while (lineIterator.hasNext()) {
             String line = lineIterator.next().trim();
             if (line.startsWith("#text=")) {
-                text.append(line.substring(6) + "\n");
+                text.append(line.substring(6)).append("\n");
                 textFound = true;
                 continue;
             }
@@ -279,7 +279,7 @@ public class WebannoTsv1Reader
                 String token = lineTk.nextToken();
 
                 // for backward compatibility
-                tmpText.append(token + " ");
+                tmpText.append(token).append(" ");
 
                 tokens.put(tokenNumber, token);
                 lemma.put(tokenNumber, lineTk.nextToken());
--- a/webanno-io-tsv/src/main/java/de/tudarmstadt/ukp/clarin/webanno/tsv/WebannoTsv3Reader.java
+++ b/webanno-io-tsv/src/main/java/de/tudarmstadt/ukp/clarin/webanno/tsv/WebannoTsv3Reader.java
@@ -153,7 +153,7 @@ public class WebannoTsv3Reader extends JCasResourceCollectionReader_ImplBase {
 				if (sentLineSb.toString().isEmpty()) {
 					sentLineSb.append(text);
 				} else {
-					sentLineSb.append(LF + text);
+					sentLineSb.append(LF).append(text);
 				}
 				lastSent = sentLineSb.toString();
 				continue;
@@ -669,13 +669,13 @@ public class WebannoTsv3Reader extends JCasResourceCollectionReader_ImplBase {
 	    
 		if (aPrevEnd + 1 < aBegin) {
 		    // FIXME This is very slow. Better use StringUtils.repeat()		    
-			String pad = ""; // if there is plenty of spaces between sentences
+			StringBuilder pad = new StringBuilder(); // if there is plenty of spaces between sentences
 			for (int i = aPrevEnd + 1; i < aBegin; i++) {
-				pad = pad + " ";
+				pad.append(" ");
 			}
-			coveredText.append(pad + aLine + LF);
+			coveredText.append(pad).append(aLine).append(LF);
 		} else {
-			coveredText.append(aLine + LF);
+			coveredText.append(aLine).append(LF);
 		}
 		Sentence sentence = new Sentence(aJCas, aBegin, aEnd);
 		sentence.addToIndexes();
--- a/webanno-api-annotation/src/main/java/de/tudarmstadt/ukp/clarin/webanno/api/annotation/rendering/RelationRenderer.java
+++ b/webanno-api-annotation/src/main/java/de/tudarmstadt/ukp/clarin/webanno/api/annotation/rendering/RelationRenderer.java
@@ -151,11 +151,11 @@ public class RelationRenderer
                 end = selectByAddr(aJCas, depFs).getEnd();
             }
             else if (end + 1 != selectByAddr(aJCas, depFs).getBegin()) {
-                cm.append(" ... " + selectByAddr(aJCas, depFs).getCoveredText());
+                cm.append(" ... ").append(selectByAddr(aJCas, depFs).getCoveredText());
                 end = selectByAddr(aJCas, depFs).getEnd();
             }
             else {
-                cm.append(" " + selectByAddr(aJCas, depFs).getCoveredText());
+                cm.append(" ").append(selectByAddr(aJCas, depFs).getCoveredText());
                 end = selectByAddr(aJCas, depFs).getEnd();
             }
 
--- a/webanno-api-dao/src/main/java/de/tudarmstadt/ukp/clarin/webanno/api/dao/CasStorageServiceImpl.java
+++ b/webanno-api-dao/src/main/java/de/tudarmstadt/ukp/clarin/webanno/api/dao/CasStorageServiceImpl.java
@@ -115,11 +115,12 @@ public class CasStorageServiceImpl
           }
           catch (CasDoctorException e) {
               StringBuilder detailMsg = new StringBuilder();
-              detailMsg.append("CAS Doctor found problems for user ["
-                      + aUserName + "] in source document [" + aDocumentName + "] ("
-                      + aDocumentId + ") in project["
-                      + aProject.getName() + "] ("
-                      + aProject.getId() + ")\n");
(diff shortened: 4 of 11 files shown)


webanno-api-annotation/src/main/java/de/tudarmstadt/ukp/clarin/webanno/api/annotation/rendering/model/VSpan.java
```diff
@@ -40,20 +41,20 @@
     public VSpan(AnnotationLayer aLayer, AnnotationFS aFS, String aType, VRange aOffsets,
             Map<String, String> aFeatures)
     {
-        this(aLayer, new VID(getAddr(aFS)), aType, asList(aOffsets), aFeatures);
+        this(aLayer, new VID(getAddr(aFS)), aType, Collections.singletonList(aOffsets), aFeatures);
```
reckart: Why is `Collections.singletonList` better than `(Collections.)asList`? asList is at least shorter ;)

webanno-api-annotation/src/main/java/de/tudarmstadt/ukp/clarin/webanno/api/annotation/rendering/model/VSpan.java
```diff
@@ -40,20 +41,20 @@
...
     }
     
     public VSpan(AnnotationLayer aLayer, AnnotationFS aFS, String aType, VRange aOffsets,
             int aEquivalenceClass, Map<String, String> aFeatures)
     {
         super(aLayer, new VID(getAddr(aFS)), aType, aEquivalenceClass, aFeatures);
-        ranges = asList(aOffsets);
+        ranges = Collections.singletonList(aOffsets);
```
reckart: See comment above.

webanno-api-annotation/src/main/java/de/tudarmstadt/ukp/clarin/webanno/api/annotation/rendering/model/VSpan.java
```diff
@@ -40,20 +41,20 @@
...
    }
 
     public VSpan(AnnotationLayer aLayer, VID aVid, String aType, VRange aOffsets,
             Map<String, String> aFeatures)
     {
-        this(aLayer, aVid, aType, asList(aOffsets), aFeatures);
+        this(aLayer, aVid, aType, Collections.singletonList(aOffsets), aFeatures);
```
reckart: See comment above.

webanno-brat/src/main/java/de/tudarmstadt/ukp/clarin/webanno/brat/render/model/Entity.java
```diff
@@ -53,7 +54,7 @@ public Entity()
 
     public Entity(int aId, String aType, Offsets aOffsets, String aLabelText, String aColor)
     {
-        this(aId, aType, asList(aOffsets), aLabelText, aColor);
+        this(aId, aType, Collections.singletonList(aOffsets), aLabelText, aColor);
```
reckart: See comment above.

webanno-brat/src/main/java/de/tudarmstadt/ukp/clarin/webanno/brat/render/model/Entity.java
```diff
@@ -63,7 +64,7 @@ public Entity(int aId, String aType, List<Offsets> aOffsets, String aLabelText,
 
     public Entity(VID aVid, String aType, Offsets aOffsets, String aLabelText, String aColor)
     {
-        this(aVid, aType, asList(aOffsets), aLabelText, aColor);
+        this(aVid, aType, Collections.singletonList(aOffsets), aLabelText, aColor);
```
reckart: See comment above.

webanno-brat/src/main/java/de/tudarmstadt/ukp/clarin/webanno/brat/render/model/EntityType.java
```diff
@@ -62,7 +63,7 @@ public EntityType(String aName, String aLabel, String aType)
     {
         this(aName /* name */, aType /* type */, true /* unused */, "" /* hotkey */,
                 null /* fgColor */, null /* bgColor */, null /* borderColor */,
-                Arrays.asList(aLabel) /* labels */, null /* children */, null /* attributes */, 
+            Collections.singletonList(aLabel) /* labels */, null /* children */, null /* attributes */,
```
reckart: Again, I'd go with `asList` and a static import.

webanno-brat/src/main/java/de/tudarmstadt/ukp/clarin/webanno/brat/render/model/RelationType.java
```diff
@@ -66,7 +67,7 @@ public RelationType()
     public RelationType(String aName, String aLabel, String aType, String aTarget, String aColor,
             String aArrowHead, String aDashArray)
     {
-        this(aColor, aArrowHead, Arrays.asList(aLabel), aType, Arrays.asList(aTarget), aDashArray);
+        this(aColor, aArrowHead, Collections.singletonList(aLabel), aType, Collections.singletonList(aTarget), aDashArray);
```
reckart: See comment above.

webanno-io-tsv/src/main/java/de/tudarmstadt/ukp/clarin/webanno/tsv/WebannoTsv3Reader.java
```diff
@@ -669,13 +669,13 @@ private void createSentence(JCas aJCas, String aLine, int aBegin, int aEnd, int
 	    
 		if (aPrevEnd + 1 < aBegin) {
 		    // FIXME This is very slow. Better use StringUtils.repeat()		    
-			String pad = ""; // if there is plenty of spaces between sentences
+			StringBuilder pad = new StringBuilder(); // if there is plenty of spaces between sentences
```
reckart: Good one :)

webanno-ui-curation/src/main/java/de/tudarmstadt/ukp/clarin/webanno/ui/curation/component/SuggestionViewPanel.java
```diff
@@ -618,7 +618,7 @@ private void addSuggestionColor(Project aProject, Mode aMode, Map<String, JCas>
             Collection<ConfigurationSet> aCfgSet, boolean aI, boolean aAgree)
     {
         for (ConfigurationSet cs : aCfgSet) {
-            boolean use = false;
+            boolean use;
```
reckart: I prefer having explicit initalizers for primitive variables, even if they initialize to the same value that the JVM would use.

webanno-io-tsv/src/main/java/de/tudarmstadt/ukp/clarin/webanno/tsv/WebannoTsv1Reader.java
```diff
@@ -146,7 +146,7 @@ private void createDependency(JCas aJCas, Map<Integer, String> tokens,
                 // if span A has (start,end)= (20, 26) and B has (start,end)= (30, 36)
                 // arc drawn from A to B, dependency will have (start, end) = (20, 36)
                 // arc drawn from B to A, still dependency will have (start, end) = (20, 36)
-                int begin = 0, end = 0;
+                int begin, end;
```
reckart: I prefer having explicit initalizers for primitive variables, even if they initialize to the same value that the JVM would use.

webanno-io-tsv/src/main/java/de/tudarmstadt/ukp/clarin/webanno/tsv/WebannoTsv1Reader.java
```diff
@@ -101,7 +101,7 @@ private void createToken(JCas aJCas, StringBuilder text, Map<Integer, String> to
             Map<Integer, String> pos, Map<Integer, String> lemma, Map<String, Token> tokensStored)
     {
         int tokenBeginPosition = 0;
-        int tokenEndPosition = 0;
+        int tokenEndPosition;
```
reckart: I prefer having explicit initalizers for primitive variables, even if they initialize to the same value that the JVM would use.

webanno-api-annotation/src/main/java/de/tudarmstadt/ukp/clarin/webanno/api/annotation/rendering/model/VSpan.java
astier: _Arrays.asList_ with just a single argument wastes memory and is less performant than _Collections.singletonList_ because singletonList is not backed by an array and just has a reference to that one item. However, its may not worth it if you say the readability suffers disproportional. Should I revert it?

reckart: I have had a look at the implementations of `asList()` and `singletonList()`. Actually, I wasn't aware that `asList()` internally simply calls `new ArrayList<>(...)`. I thought it would return a non-modifiable list. Well, always a good thing to learn. In fact, I was wishfully thinking that `asList(...)` may internally even switch between a multi-element and a single-element list implementation depending on the number of elements passed to it. I am not sure if (IntelliJ is smart enough to detect if) the switch from `asList()` to `singletonList()` safe - one is modifiable, the other is not. We could bulk-switch to`singletonList()` and fix those cases were we might get exceptions then (although I may have a hard time switching my head do that). I am a bit ambivalent... What do you think about the risk of getting exceptions due to the switch?

astier:
I adjusted the changes. I removed also the singleton stuff. I think right now the risk isn't worth the benefit.  Especially because I don't want to be responsible for weird bugs.

Besides that the PR should be ready.

reckart:
Thanks :)

Issue #531, opened 2017-06-16
adaptive coloring strategy


remstef:
refactor selection of coloring strategy, and make it independent of "readonly" option

E.g. would the following mockup be an option?

![adaptive colering strategy](https://user-images.githubusercontent.com/5611665/27218209-75b40cca-527d-11e7-8e01-8f078330246e.png)

reckart:
Hm... IMHO visibility and coloring are orthogonal. I.e. when I "unhide" a layer, it should IMHO show again using the previously configured coloring scheme.

What made you consider the per-layer coloring again? I though last time we talked, we had gravitated towards having options in the lower part like:

- Use the same color for all tags in a layer y/n
- Colorize non-editable annotations y/n

remstef:
>  when I "unhide" a layer, it should IMHO show again using the previously configured coloring scheme

that's a fair point, I agree to that 

I think we need that in a way configurable per layer, we are restricting ourselves too much if its a global option. E.g. unchecking the "Use the same color for all tags" might make sense when working with specific tagsets like POS tags, but it does not really make sense when using free text. Another possibility is to put that kind of option into the layer configuration in the project page, only admins can then change it, in order not to overwhelm the user. What's your take on that?

reckart:
Since different users perceive colors differently, I think it makes more sense to have color-related settings close to the user - not close to the admin.

Ok, so let's say we make the label-based or layer-based coloring strategy something that can be configured per layer - a dropdown would seem useful for that.

To me the question whether a read-only layer should be greyed out or not seems kind of independent of the chosen strategy though. That is, I tend towards believing that if a user wants read-only layers to be colored, it should apply to all read-only layers - likewise if the user wants them to be monochrome or possibly even transparent.

What do you think about allowing to configure the coloring-strategy via  dropdown per layer and having another global dropdown labeled "Colorize read-only layers" with the options "normal" and "gray-scale"?

remstef:
That sounds good to me.

remstef:
I'm wondering where to store the preferred coloring strategy. On the one hand it is similar `AnnotationPreference.isStaticColor`, but on the other hand it is similar to `AnnotationLayer.isEnabled` and a list of `AnnotationLayer` objects is also stored in `AnnotationPreference`. @reckart do you have an opinion?

reckart:
Since the preference is per-user, it has to be stored in AnnotationPreference.

remstef:
[...]

PR #544, opened 2017-06-23
#531 - adaptive coloring strategy

remstef:
- select coloring strategy per layer in settings

### DIFF
--- a/webanno-api-annotation/src/main/java/de/tudarmstadt/ukp/clarin/webanno/api/annotation/model/AnnotationPreference.java
+++ b/webanno-api-annotation/src/main/java/de/tudarmstadt/ukp/clarin/webanno/api/annotation/model/AnnotationPreference.java
@@ -56,7 +56,8 @@ public class AnnotationPreference
     
 //    // determine if static color for annotations will be used or we shall
 //    // dynamically generate one
-    private boolean staticColor = true;
+    @Deprecated 
+    private boolean staticColor = true; // this is only here to not break previous user settings
     
 	private Map<Long, ColoringStrategyType> colorPerLayer;
     
@@ -160,11 +161,13 @@ public class AnnotationPreference
 		this.colorPerLayer = colorPerLayer;
 	}
 
+	@Deprecated 
     public boolean isStaticColor()
     {
         return staticColor;
     }
 
+    @Deprecated
     public void setStaticColor(boolean staticColor)
     {
         this.staticColor = staticColor;
--- a/webanno-api-annotation/src/main/java/de/tudarmstadt/ukp/clarin/webanno/api/annotation/model/AnnotationPreference.java
+++ b/webanno-api-annotation/src/main/java/de/tudarmstadt/ukp/clarin/webanno/api/annotation/model/AnnotationPreference.java
@@ -22,6 +22,7 @@ import java.util.List;
 import java.util.Map;
 
 import de.tudarmstadt.ukp.clarin.webanno.api.annotation.coloring.ColoringStrategy.ColoringStrategyType;
+import de.tudarmstadt.ukp.clarin.webanno.api.annotation.coloring.ColoringStrategy.ReadonlyColoringBehaviour;
 
 /**
  * This is a class representing the bean objects to store users preference of annotation settings
@@ -57,9 +58,11 @@ public class AnnotationPreference
 //    // determine if static color for annotations will be used or we shall
 //    // dynamically generate one
     @Deprecated 
-    private boolean staticColor = true; // this is only here to not break previous user settings
+    private boolean staticColor = true; // this is only here to not break previous user settings, its not an option that can be set anymore
     
 	private Map<Long, ColoringStrategyType> colorPerLayer;
+	
+	private ReadonlyColoringBehaviour readonlyLayerColoringBehaviour = ReadonlyColoringBehaviour.LEGACY;
     
     private int sidebarSize;
     private int fontSize;
@@ -159,6 +162,16 @@ public class AnnotationPreference
 	public void setColorPerLayer(Map<Long, ColoringStrategyType> colorPerLayer) 
 	{
 		this.colorPerLayer = colorPerLayer;
+	}
+	
+	public ReadonlyColoringBehaviour getReadonlyLayerColoringBehaviour() 
+	{
+		return readonlyLayerColoringBehaviour;
+	}
+
+	public void setReadonlyLayerColoringBehaviour(ReadonlyColoringBehaviour readonlyLayerColoringBehaviour) 
+	{
+		this.readonlyLayerColoringBehaviour = readonlyLayerColoringBehaviour;
 	}
 
 	@Deprecated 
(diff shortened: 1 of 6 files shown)


reckart:
@remstef lots of changes ;) Could you please make one more and apply the usual code style - there is an [Eclipse style file](http://dkpro.github.io/files/DKProCoreStyle_20120326.xml).

reckart:
Please apply it mainly to new files and blocks where you did major changes. E.g. it seems you are using tabs instead of spaces, a different style of placing curly braces, etc. Please don't always globally apply the style to all edited files on save or even to the whole project since that may lead excessive changes in whitespace unrelated to the current task.

reckart:
@remstef Looking at ColoringStrategy.java, it seems that actually some methods did not change at all. GitHub diff just doesn't get it because the indentation/formatting style change. Can you please undo all changes that are whitespace only?

webanno-api-annotation/src/main/java/de/tudarmstadt/ukp/clarin/webanno/api/annotation/model/AnnotationPreference.java
```diff
@@ -119,21 +128,59 @@ public void setScrollPage(boolean aScrollPage)
         scrollPage = aScrollPage;
     }
 
+    /**
```
reckart: This type of empty JavaDoc doesn't serve any purpose. Could you please remove it (unless of course you would like to write proper JavaDoc ;) )

webanno-api-annotation/src/main/java/de/tudarmstadt/ukp/clarin/webanno/api/annotation/model/AnnotationPreference.java
```diff
@@ -119,21 +128,59 @@ public void setScrollPage(boolean aScrollPage)
...
+		this.readonlyLayerColoringBehaviour = readonlyLayerColoringBehaviour;
+	}
+
+	@Deprecated 
     public boolean isStaticColor()
     {
         return staticColor;
     }
 
+    @Deprecated
```
reckart: Is it still called from anywhere? Couldn't it be removed?

webanno-ui-annotation/src/main/java/de/tudarmstadt/ukp/clarin/webanno/ui/annotation/PreferencesUtil.java
```diff
@@ -114,6 +125,14 @@ public static void loadPreferences(String aUsername, SettingsService aSettingsSe
                         .getProject());
                 aBModel.setAnnotationLayers(layers);
             }
+
+            // Get color preferences for each layer, init with legacy if not found
```
reckart: Did you try what happens if `(preference.getColorPerLayer() != null` and an admin adds a new layer? Will there be an exception? What coloring strategy will it have?

webanno-ui-annotation/src/main/java/de/tudarmstadt/ukp/clarin/webanno/ui/annotation/dialog/AnnotationPreferenceModalPanel.html
```diff
@@ -65,7 +92,7 @@
 			</tr>
 			<tr class="border_bottom">
 				<td><label> Use the same color for all tags in a layer </label></td>
```
reckart: I this label `Use the same color for all tags in a layer` still correct after the checkbox was changed to `readonlylayercoloring`?

webanno-api-annotation/src/main/java/de/tudarmstadt/ukp/clarin/webanno/api/annotation/model/AnnotationPreference.java
remstef: it could be removed actually

webanno-ui-annotation/src/main/java/de/tudarmstadt/ukp/clarin/webanno/ui/annotation/PreferencesUtil.java
remstef: ok, it doesn't occur, I have to repair that

webanno-ui-annotation/src/main/java/de/tudarmstadt/ukp/clarin/webanno/ui/annotation/dialog/AnnotationPreferenceModalPanel.html
remstef: lol, no... I have missed that one :)

remstef:
regarding style and whitespaces, just to be on the save side, is it enough to for if I apply the style file to every file I touched or should make a commit that reverts everything and then another that has the changes with the correct style and indentation?

reckart:
@remstef wrt style:  Look at the diff of the commit and you will see that especially for the two files that I mentioned, basically the whole file appears to be changed although you only changed certain parts of it. Please revert any changes to whitespace outside the blocks of code that you touched *within* a file. Please apply the style only to changed *blocks* - not entire files - unless the files are completely new.

webanno-ui-annotation/src/main/java/de/tudarmstadt/ukp/clarin/webanno/ui/annotation/PreferencesUtil.java
remstef: last commit b20a5fd should account for that, at least it does during debugging. But it still doesn't show up in preferences because I'm iterating over `bModel.getAnnotationLayers()` and that doesn't seem to get the change. Might I suggest exchanging `List<Long> getAnnotationLayers()` in  `AnnotationPreference` with stg like `Map<Long, Boolean> getAnnotationLayerVisibility()`.

reckart: As I said in https://github.com/webanno/webanno/issues/531#issuecomment-310433982 and https://github.com/webanno/webanno/issues/531#issuecomment-310436267 : it might not be the worst idea to change way that the preferences are being saved from the present "manual" way to using Jackson along with changing the datamodel. In such a case, I would probably go with `Map<Long, LayerPreferences>`.

reckart:
I have set up a checkstyle config to reduce style deviations in the future. Unfortunately, that caused some conflicts in this PR, but since 99% of the checkstyle-related changes are whitespace, these conflicts should be easy to resolve.

webanno-ui-annotation/src/main/java/de/tudarmstadt/ukp/clarin/webanno/ui/annotation/PreferencesUtil.java
remstef: `Map<Long, LayerPreferences>` seems reasonable, I'd prefer that in a separate issue though

> As I said in ...

Yess, and now I also know much better why you mentioned that. I couldn't really follow before ;)

reckart:
Since you are working on a feature branch, I would tend towards rebasing changes in the branch to the latest master instead of merging master into the feature branch with a merge commit. What do you think?

remstef:
I guess that makes sense. Rebase keeps the history cleaner, right? I never really worked with rebase but I will try that next time.

reckart:
At least the merges don't appear. It seems to be a cultural choice - some prefer the detailed history with all the merge info - others prefer a cleaner history. The same applies when merging feature branches into master - some merge, some rebase, some squash. I personally tend towards a cleaner history - but I'm also still learning how to get most out of git.

Issue #520, opened 2017-06-08
Perform javascript action on click


remstef:
Allow onClick events on spanannotations and run a defined Javascript action.

reckart:
Do we need the ability to put arbitrary JS there or wouldn't an URL-pattern that could be constructed from a fixed set of parameters (e.g. documentId, projectId, userId, annotation features) be sufficient. I think a URL-pattern would be easier for the typical to understand/configure and less error-prone.

remstef:
Javascript is more flexible in my opinion, e.g. we need to open a particular URL always in the same window/tab. When we supply only a URL pattern, we'd also need an option for a window name and so on. Also this helps a lot for debugging and offers a lot of flexibility.

reckart:
If you want to use JS, then I think the string-substitution in the code-snippet is not a good solution. It introduces special meaning to certain characters not present in JS. For a JS-based solution, it would seem more sensible to pass the parameters as a dictionary, e.g.

```
function handleAction(parameters) {
  // this is where the user-supplied snippet goes and where it can access the "parameters" variable.
}
```

reckart:
Also, I think "doAction" should be renamed to "onClick" then - next you may want to override "onMouseEnter" or something else. Being aligned with the event names in JS would be helpful then.

remstef:
I agree, I named this option "doAction" to be in line with the other options created by other events, e.g. in onDblClick: "createArc", "createSpan"

reckart:
It depends whether taking a Annotation-centric perspective or a JS-centric. I had the impression you were taking an JS-centric perspective since you were aiming for maximal flexibility.

remstef:
> If you want to use JS, then I think the string-substitution in the code-snippet is not a good solution. 

I think the template option is the simplest, we don't want complex Javascript functions here, only things that could be written in a single line, max 2 or three commands. It is still possible to call more complex Javascript functions, but users can then build their parameter objects here too. 

> It depends whether taking a Annotation-centric perspective or a JS-centric.

In the end it's a click on an annotation. But I really don't mind renaming this to "onClick", I'd follow your lead here.

reckart:
[...]

PR #549, opened 2017-06-26
#520 - Perform javascript action on click

remstef:
- increase column size

### DIFF
--- a/webanno-model/src/main/java/de/tudarmstadt/ukp/clarin/webanno/model/AnnotationLayer.java
+++ b/webanno-model/src/main/java/de/tudarmstadt/ukp/clarin/webanno/model/AnnotationLayer.java
@@ -71,7 +71,8 @@ public class AnnotationLayer
     
     private boolean readonly = false;
     
-    @Column(nullable = true, length = 128000)
+    @Lob
+    @Column(nullable = true, length = 64000)
     private String onClickJavascriptAction;
 
     @Column(name = "name", nullable = false)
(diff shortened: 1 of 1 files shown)


webanno-model/src/main/java/de/tudarmstadt/ukp/clarin/webanno/model/AnnotationLayer.java
```diff
@@ -71,7 +71,7 @@
     
     private boolean readonly = false;
     
-    @Column(nullable = true)
+    @Column(nullable = true, length = 128000)
```
reckart: This is what we normally use for such columns. Would 64kb be ok for you?

```
    @Lob
    @Column(length = 64000)
```

remstef: yes, should be sufficient :) 
what's `@Lob`?

reckart: Marks the column as a "large binary object". As far as I understand, fepending on the DB, such columns are not stored directly in the table, but rather the table contains pointers into a separate LOB storage area. Afaik that is meant to reduce the data that is kept in memory when the table are loaded/scanned.

reckart: https://docs.jboss.org/hibernate/jpa/2.1/api/javax/persistence/Lob.html

remstef:
Alright, I changed the definition accordingly

PR #560, opened 2017-07-03

astier:
- Refactor TextFeatureEditor constructor
- Fix misspellings
- Change access modifier


astier:
Can you please merge-squash this time? I would like to try it out.

reckart:
Done. Did it work as expected?

astier:
Yes it worked very good. Was it complicated for you? I do not really like my other approach where I squash and rebase the branch because then all the older commits and our PR conversation get lost. The git book also recommends either that the maintainer merge(-squashes) the original branch or if one wants to rebase the branch then this should be done with a second identical branch. For the second rebased branch a second PR should be opened which references to the original PR, so that the original conversation- and commit-history gets preserved. In all my old PRs where I squashed and rebased its not clear any more what was the history of the PR.

What do you think. Is it ok for you when in the future I don't squash and rebase but instead you merge-squash?

reckart:
Correct me, but if I do the merge-squash, then the commit history also is discarded, no? I mean, that's the effect of the squash. It would be nice to preserve the commit history - you only need to write properly formatted commit comments - then I would just to a rebase to integrate the PR.

astier:
Yes, the commit history gets lost in the webanno history. So in the webanno/master branch one will see only one squashed commit with all the changes from the PR. However, what I am talking about is that the history in the PR should stay. One always could go back to the PR and see what happened there originally even tho its all squashed to one commit in the webanno/master branch.

The problem I have with your suggestion is that then I have to call every single commit _#<nr> - issue headline_ instead of directly saying what I did in this single atomic commit. It gets hard to review what happened by only locking at the commit history. One will always have to click on a specific commit to see the further description what happened in there. Also it gets very meaningless if you look at the history of the PR. Well, because every commit has the same headline.

The approach I suggest would have the benefit that in the webanno/master branch there won't be 20 commits with the same headline which all belong together. Instead it would be only one. On the other side the exact history is still remained in the PR with atomic commits which have meaningful names.

But of course I will handle it as you tell me, based on how you would like to keep your PRs and the master branch history.

reckart:
I don't mind doing the squash/merge if you prefer that.

PR #581, opened 2017-07-20
Bug when creating a layer with a name from the dropdown

astier:
When one wants to create a new layer and clicks in the layer-name-field and selects from the
drop-down-list, which shows the previously typed in layer-names, an exception gets thrown or the application complains that this layer already exists although it doesn't.

The problem is that the the layer-name-field only gets triggered on an keyup-  instead on an input-event.


astier:
Haha this PR history is messy.

reckart:
Not as messy as the code it tries to patch ;)

astier:
When you're happy with this PR you can merge.

Issue #618, opened 2017-08-15
Delete multiple documents at once


astier:
### Problem
Currently it is only possible to delete one document at a time from a project.
In a project with a huge number of documents this can become very tedious.

### Feature Request
It would be nice if one could select all or multiple documents which should be deleted.

reckart:
Would you like to do a PR, then we might still be able to fit it into 3.3.0?

PR #619, opened 2017-08-15
#618 - Enable multiple select on document list in Project config page

kazukisan:
#618 Enable multiple selection for documents list in project settings view.


reckart:
Thanks! :) I tried it. If that simple change would have worked, it would have been fascinating! ;) Although the browser indeed allowed selecting multiple items, the backend only deleted the first one.

I pushed a fix to #618 which allows multi-selection and also updates the confirmation dialog to show the number of files to be deleted.

kazukisan:
Thanks for the fix!

Issue #667, opened 2017-10-06
Annotations starting within token A and ending within token B not supported


reckart:
The TSV3 format presently does not seem to support annotations which end in the middle of token B. E.g.

```
[This i]s
T[his i]s
```

* In the writer, a recursive endless loop is triggered in getSubUnits(), leading to a StackOverflowException.
* The reader treats the two subtoken fragments as separate annotations instead of belonging together, i.e. on reading,

```
#Text=This is a test .
1-1	0-4	This	*[1]	
1-2	5-7	is	_	
1-2.1	5-6	i	*[1]	
1-3	8-9	a	_	
1-4	10-14	test	_	
1-5	15-16	.	_	
```

becomes 

```
#Text=This is a test .
1-1	0-4	This	*	
1-2	5-7	is	_	
1-2.1	5-6	i	*	
1-3	8-9	a	_	
1-4	10-14	test	_	
1-5	15-16	.	_	
```

reckart:
@seyyaw  It seems to be a case that was not considered so far. I have set up unit tests and prepared a fix to the writer part in (#668).

The reader part seems more difficult. The code seems to expect that multi-token (sub)units always follow each other directly in the TSV. However, in the case that an annotation ends in the middle of a token, that is not the case:

```
#Text=This is a test .
1-1	0-4	This	*[1]	
1-2	5-7	is	_	
1-2.1	5-6	i	*[1]	
```

The reader code, thus, does not detect that 1-2.1 is still part of 1-1 and creates two annotations instead when reading this file.

reckart:
@seyyaw an option to facilitate fixing the reader might be to place prefix-subtokens *before* the corresponding main token, e.g.

```
#Text=This is a test .
1-1	0-4	This	*[1]	
1-2.1	5-6	i	*[1]	
1-2	5-7	is	_	
1-3	8-9	a	_	
1-4	10-14	test	_	
1-5	15-16	.	_	
```

WDYT?

reckart:
This works with the new reader/writer from #676. Covered by tests `testStackedSubMultiTokenSpanWithFeatureValue` and `testSubMultiTokenSpanWithFeatureValue`.

PR #668, opened 2017-10-06
#667 - Annotations starting within token A and ending within token B not supported

reckart:
Unit test and several fixes to the writer.

Fixing the reader seems to be more complicated.

### DIFF
--- a/webanno-io-tsv/pom.xml
+++ b/webanno-io-tsv/pom.xml
@@ -85,6 +85,11 @@
       <artifactId>de.tudarmstadt.ukp.dkpro.core.testing-asl</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>de.tudarmstadt.ukp.clarin.webanno</groupId>
+      <artifactId>webanno-io-xmi</artifactId>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
   <build>
     <pluginManagement>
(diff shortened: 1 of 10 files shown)


reckart:
@seyyaw There is a bug in the TSV3 reader and writer (#667). It seems to be a case that was not considered so far. I have set up unit tests and prepared a fix to the writer part.

The reader part seems more difficult. The code seems to expect that multi-token (sub)units always follow each other directly in the TSV. However, in the case that an annotation ends in the middle of a token, that is not the case:

```
#Text=This is a test .
1-1	0-4	This	*[1]	
1-2	5-7	is	_	
1-2.1	5-6	i	*[1]	
```

The reader code, thus, does not detect that 1-2.1 is still part of 1-1 and creates two annotations instead when reading this file.

seyyaw:
@reckart Ok: I will check it

reckart:
@seyyaw if you want to make any code changes related to this, please commit them to the `issue667-tsv3multisubtoken` branch, not to master or 3.3.x.

seyyaw:
Ok good

reckart:
@seyyaw I have added additional unit tests for complex combinations of tokens, subtokens and zero-witdh tokens. It seems that there are still some issues even in writing the data. E.g. in one case, a subtoken that was in the CAS does not appear at all in the TSV3 output. However, most problems appear to be in the reading of the data. It seems that reading of sub-token annotation is really only supported in a few very simple cases. 

So we have various failing tests now, some with preliminary reference data, some without any reference data yet. But the tests can be used for debugging / fixing.

reckart:
This works with the new reader/writer from #676. No need anymore for this PR.

Issue #710, opened 2017-11-16
Tsv3XCasDocumentBuilder creates negative spans for zero-length annotations


adrianeboyd:
The ``beginToken`` for a zero-length annotation in Tsv3XCasDocumentBuilder seems to be the previous token, so it ends up creating a negative span with min() below and then Tsv3XSerializer throws an exception due to a negative substring.

```
TsvSubToken t = beginToken.createSubToken(annotation.getBegin(), 
    min(beginToken.getEnd(), annotation.getEnd()));
```

reckart:
This is a duplicate of #713.

PR #711, opened 2017-11-16
 #710 Tsv3XCasDocumentBuilder creates negative spans for zero-length …

adrianeboyd:
…annotations

- change zero-length token end to end of annotation unit


reckart:

reckart:
I have fixed the issue (#710 == #713) in a slightly different way by enforcing that for zero-width annotations, the begin and the end token are always ensured to be the same.

reckart:
Thanks again for the reports and the investigation! This was really a great catch!

Issue #316, opened 2016-07-20
Text selection behavior while dragging mouse


kazukisan:
I am interested in a task to highlight a region of text (a paragraph or more at a time), giving it a label, etc. It is kind of like topic segmentation annotation in a free form text.

I noticed that in WebAnno (and BRAT), when I try to select a range of text, starting by first clicking and dragging the mouse, when the mouse moves to blank text region (often blank lines between paragraphs), the highlighting of the text done so far gets removed, but it highlights text from the initial drag position back to the beginning. If I move the mouse further into the next paragraph, it will highlight the appropriate amount of text region again.

The issue is the flickering behavior, in which the highlighted region alternates between the true selected region versus highlighting from the beginning of the text to the initial drag start position.

This issue doesn't happen in typical websites like wikipedia and most other sites. The highlighted regions in these are from the drag start position X/Y to the current mouse dragged position X/Y.

reckart:
I did a brief investigation on this:

* applying the CSS class "unselectable" to the background elements in the SVG doesn't help. I had some hopes that making the background unselectable might solve the problem... but no. Maybe I did not try everything possible...
* brat has the same problem, in particular when moving the mouse between lines. Within a line, brat tries to alleviate the problem by appending a single space character to the SVG tspan elements. However, when the rendering adds extra spacing between words to make room for the labels/arcs, the flashing effect becomes more prevalent.

For intra-line flashing, it might work to add a whitespace tspan with manually set textLength attribute between the word tspans... but for the inter-line flashing, I still have no idea.

GiantEnemyCrab:
Thank you looking into this @reckart .

I did some investigation myself, too, and have a work-in-progress version.
I made a screen recording of it (in BRAT actually) comparing before vs after.

https://www.youtube.com/watch?v=VHBicsMbgWM

I am using getSelection()'s range objects to overwrite the text selection behavior, but highlighting is snapped to token-level, instead of character-level. I would want to do this at a character level. I haven't figured out how to do this at a character-level yet, but this does address the issues I mentioned. Let me know how that video looks to you.

reckart:
@GiantEnemyCrab youtube tells me the video has been removed by the user.

GiantEnemyCrab:
Sorry can you refresh this page, I updated the link.

reckart:
[...]

PR #715, opened 2017-11-17
#316 - Text selection behavior while dragging mouse

reckart:
- Create selectable stretching spacer element between chunks
- Adjust annotation creation code to handle spacer elements

### DIFF
--- a/webanno-brat/src/main/java/de/tudarmstadt/ukp/clarin/webanno/brat/resource/annotator_ui.js
+++ b/webanno-brat/src/main/java/de/tudarmstadt/ukp/clarin/webanno/brat/resource/annotator_ui.js
@@ -1723,9 +1723,25 @@ var AnnotatorUI = (function($, window, undefined) {
         } else if (!evt.ctrlKey) {
           // if not, then is it span selection? (ctrl key cancels)
           var sel = window.getSelection();
+          
+// BEGIN WEBANNO EXTENSION - #316 Text selection behavior while dragging mouse
+/*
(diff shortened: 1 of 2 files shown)


kazukisan:
Thanks for the PR on this. But on Mac OSX, I somehow still experience the same behavior with regard to jittery highlighting behevior on both Chrome  (ver 62) and Safari (ver 10.1.2).

What is your environment like to confirm the fix at least on some platforms? (I think I followed the correct way to do this, ```git branch``` shows feature/issue316-smooth-selection and locally, the application is from the fresh mvn clean installed.

reckart:
The video below was recorded on OS X with Safari Version 11.0.1 (11604.3.5.1.1). When dragging within a single row the jitter is no longer there. Before, in particular the empty areas between the tokens would cause jitter.



However, there is still jitter when dragging across rows.



If you don't get the effect from the first video, maybe you have to clear your browser cache to ensure that you use the latest rendering code.

kazukisan:
Thanks for the clarification!
I was testing for "across-row" highlighting, but the the screen recording makes it clear what improved. Thanks!

reckart:
@kazukisan I had a look again at the screen recording for the changes you made to improve the selection process (https://www.youtube.com/watch?v=uLy8TaHbWE8), but I don't see very well cross-row effects. Can you tell if cross-row works better in your approach than in mine?

kazukisan:
The approach I took for making cross-row effect good was to make the text highlighting work on character-level, It was only half-done in its implementation (not as optimized for performance), maybe something to look into again or from a different perspective. I just know that the code needs to be adjusted to work with the current WebAnno code.

Issue #724, opened 2017-11-19
Cross-row selection is jumpy


reckart:
While the changes for #316 fix in-row selection jumpyness, the cross-row selection jumpyness still exists in LTR as well as RTL mode and on all browsers.

![chrome_error-firefox](https://user-images.githubusercontent.com/19265315/32554122-20320a48-c499-11e7-940d-cf6f522b3a1f.gif)

See also:
* #316 - Text selection behavior while dragging mouse 
* #699 - Chrome / Chromium and span annotations result in error

PR #737, opened 2017-11-25
#724 - Cross-row selection is jumpy

reckart:
- Add spacers between rows
- Add spacers at beginning and end of rows
- Always render using SVG text elements, even in LTR mode to avoid maintaining two different rendering branches
- Disable pointer events on unrelevant areas during text selection

### DIFF
--- a/webanno-brat/src/main/java/de/tudarmstadt/ukp/clarin/webanno/brat/resource/annotator_ui.js
+++ b/webanno-brat/src/main/java/de/tudarmstadt/ukp/clarin/webanno/brat/resource/annotator_ui.js
@@ -360,6 +360,16 @@ var AnnotatorUI = (function($, window, undefined) {
           startArcDrag(id);
           return false;
         }
+// BEGIN WEBANNO EXTENSION - #724 - Cross-row selection is jumpy
+        // If user starts selecting text, suppress all pointer events on annotations to
+        // avoid the selection jumping around. During selection, we don't need the annotations
+        // to react on mouse events anyway.
+        if (target.attr('data-chunk-id')) {
+          $(svgElement).children('.row, .sentnum').each(function(index, row) {
+            $(row).css('pointer-events', 'none');
+          });
+        }
+// END WEBANNO EXTENSION - #724 - Cross-row selection is jumpy
       };
 
       var onMouseMove = function(evt) {
@@ -1678,6 +1688,15 @@ var AnnotatorUI = (function($, window, undefined) {
       var onMouseUp = function(evt) {
         if (that.user === null) return;
 
+// BEGIN WEBANNO EXTENSION - #724 - Cross-row selection is jumpy
+        // Restore pointer events on annotations
+        if (dragStartedAt && $(dragStartedAt.target).attr('data-chunk-id')) {
+          $(svgElement).children('.row, .sentnum').each(function(index, row) {
+            $(row).css('pointer-events', 'auto');
+          });
+        }
+// END WEBANNO EXTENSION - #724 - Cross-row selection is jumpy
+
         var target = $(evt.target);
 
         // three things that are clickable in SVG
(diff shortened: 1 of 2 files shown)


kazukisan:
Thanks for looking into this!

I tried and it is smooth without the flickering behavior in Chrome. I don't see much behavior difference in Safari 10.1.2. And in Firefox, multi-line annotation is not possible any more. So I take it that it was developed with Chrome for now. Very nice though.

I am curious if the spacer after the last character in the line could be much shorter. It currently occupies the whole rest of the whitespace in the line.

And as you may have also noticed, sometimes the highlighting gets into the row number panel. It looks a bit weird but doesn't seem to be affecting the actual annotation process.

reckart:
Thanks for testing. I developed in Safari, but usually Chrome and Safari are very similar. I though that I also had tested cross-line annotation in Firefox... need to test that again.

It would probably be possible to do some fine-tuning on the spacer after the last character. The way it is currently, you can drag the mouse anywhere in the post-text space and still create an annotation. If I remember correctly, then I could make the spacer unselectable, then the selection would still be smooth, but releasing the mouse on the spacer would no longer create an annotation. Maybe I didn't dig deep enough. 

I only noticed selection spilling over into the row numbers if the selection goes beyond the bottom end of the annotation area. But it seemed to me that didn't happen a lot in normal operations. Was it annoying? Did it happen a lot for you? I tried making the row numbers unselectable, but that didn't seem to help (yet). Again, maybe it requires some extra digging.

reckart:
@kazukisan If you say you didn't notice any difference in Safari: try clearing the caches in Safari and try again. It should be as smooth in Safari as it is in Chrome. Clearing caches might also help in Firefox.

kazukisan:
Clearing or emptying cache in Firefox 57 / Safari 10.1.2 didn't seem to help see the new behavior. (Mac 10.10.5) Weird that I didn't see it in Safari though both Chrome and Safari should be both eventually webkit-derived.

And as far as the selection spilling over, it doesn't occur that much, and while I don't know what causes it, but I guess there may be some patterns?

I've tried this particular xmi file and there are lines that say "LA" and "JG" in this xmi, and that's where the spill-over happens:
https://gist.github.com/kazukisan/1b385ef2d47a9376768cc7249cd205bc

reckart:
I can reproduce the cross-line problem in Firefox.

However, in Safari Version 11.0.1 (11604.3.5.1.1) on Mac 10.11.6, the smooth selection works nicely (as in Chrome).

reckart:
I have rebased the PR onto the latest master. If you have a local clone of the branch, discard it and check it out again.

reckart:
The selected range seems to be reported by Firefox in a strange way. That needs some further investigation.

Also, Firefox still seems to jitter a bit - they seems to be some background shining through between the text and the inter-row spacers.

I didn't notice any highlighting of the sentence numbers in that XMI using Safari Version 11.0.1.

reckart:
@kazukisan Cross-line selection should now work again on Firefox. I didn't look into the remaining jitter on Firefox though.

kazukisan:
I can confirm in Firefox cross sentence annotation is possible again. Thank you for the fix.

An interesting observation, probably nothing can be done about this, but when I highlight text (and before releasing the mouse), I do "copy" shortcut to copy the text. In Chrome, for every word, new line is inserted into the copied text, whereas in Firefox, the copied text is more consistent with how it is displayed in WebAnno. (I sometimes copy and paste from WebAnno when I inform other people about how it should've been annotated, etc)

reckart:
Well, the proper thing to do would probably be to intercept the copy keystrokes and then trying to fill the copy buffer manually, i.e. directly from the document text stored in the internal datastructures of the UI code. But this would be a separate issue/feature request.

reckart:
I opened a separate issue for the copying thing: https://github.com/webanno/webanno/issues/745

Regarding the remaining jitter on Firefox: I didn't find a place where the background actually shines through. It seems that the selection process on Firefox just behaves a bit funky. It is not too bad and I don't have any idea how to address it. If it is still too annoying, somebody should open a new issue and optimally come up with an idea how to approach it.

Issue #752, opened 2017-12-04
New annotation layers start out hidden


reckart:
Check that new annotation layers do not start out hidden on the annotation/curation pages.

* Create a new custom annotation layer
* Go to the annotation page
* Go to the settings dialog
* Check that the new layer is visible

Looks like this is a bug starting with WebAnno 3.3.0.

reckart:
@remstef You had refactored the layer settings for #531 - I guess this bug was introduced as part of that refactoring.

remstef:
I don't think it's a bug, its more like a design decision, and it probably existed before. The thing is, when preferences are loaded there is no chance if distinguishing a new layer from an existing layer, and a user keeps a list of preferred layers. We could refactor the preference behavior, i.e. a user keeps a list of not-preferred (or hidden) layers, i.e. the layers she wants to hide. That will solve the new layer issue, because it is not in that list by default.

remstef:
Give me a GO, if you think it should be done.

reckart:
We had previously discussed about refactoring the whole saving/loading code for the user preferences from a hand-cobbled serialization to something more standard and robust, e.g. modelling preferences as Java objects and storing/loading them as JSON via Jackson. But that would be a larger refactoring. 

If you see a way of resolving the issue with little effort and without refactoring the whole preference storage, I'd say it should be worth it. If it is not easily doable, I'd say let's push this to 3.4.x or to a later 3.3.x release.

@remstef WDTY?

remstef:
> We had previously discussed about refactoring the whole saving/loading
yes, I remember and I agree, that is a larger thing

I think refactoring the logic of preferred layers -> hidden layers is possible with little effort. I will see to that.

remstef:
PR: #752 - New annotation layers start out hidden #764

PR #764, opened 2017-12-06
#752 - New annotation layers start out hidden

remstef:
- refactor logic from preferred layers to hidden layers, which resolves the hidden new layers startout hidden issue by assuming no layer is hidden in the beginning
- *NOTE* this is a branch from patch-755


reckart:

PR #766, opened 2017-12-07
#765 - Trigger BeforeDocumentRemovedEvent before removing annotation documents

### LINKED ISSUE #765: Trigger BeforeDocumentRemovedEvent before removing annotation documents
The DocumentService.removeSourceDocument method must trigger the BeforeDocumentRemovedEvent as its first action. This is important for methods that rely on this event, since they might need to have access to the annotation documents associated to the source document that is about to be removed.

betoboullosa:
- BeforeDocumentRemovedEvent is now the first action febore the removal


reckart:

Issue #798, opened 2018-01-03
Remote API v2 inconsistent URL pattern


kazukisan:
Delete document requires a slash at the end while all other end points don't.

```
DELETE /api/v2/projects/{projectId}/documents/{documentId}/
```

Pull request coming.

reckart:
Thanks! Btw. for these trivial changes, doing just a PR is fine. I can rewrite the commit comment and include the PR number when squash-merging the PR. For non-trivial PRs, having a separate issue is certainly preferable because there is a "risk" of having multiple commits in the PR and then rewriting the commit messages is rather tedious.

kazukisan:
Thanks for the note !

reckart:
Merged and cherry-picked into 3.3.x.

PR #799, opened 2018-01-03
#798 Remove a slash at the end from document deletion API

kazukisan:
#798

### DIFF
--- a/webanno-remote/src/main/java/de/tudarmstadt/ukp/clarin/webanno/webapp/remoteapi/RemoteApiController2.java
+++ b/webanno-remote/src/main/java/de/tudarmstadt/ukp/clarin/webanno/webapp/remoteapi/RemoteApiController2.java
@@ -634,7 +634,7 @@ public class RemoteApiController2
     @ApiOperation(value = "Delete a document from a project")
     @RequestMapping(
             value = "/" + PROJECTS + "/{" + PARAM_PROJECT_ID + "}/" + DOCUMENTS + "/{" 
-                    + PARAM_DOCUMENT_ID + "}/", 
+                    + PARAM_DOCUMENT_ID + "}", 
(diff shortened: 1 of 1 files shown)


reckart:
@kazukisan curiosity: are you actually using the master branch or are you just submitting PRs against the master branch? If you are using master, I'd be very curious if Spring 5, Hibernate 5 and Liquibase are working ok for you. If a change would be applicable for 3.3.x, it'd be great if you could submit the PR against the 3.3.x branch. I'm merging 3.3.x regularly into master.

kazukisan:
I am just submitting PR against master and I was thinking that a commit would be pulled from master to other branches in workflow that I am used to. But I am indeed using 3.3.x branch because I am not able to get master branch working in my environment. I will keep in mind about that.

reckart:
(Probably) I am not following any particular best-practice for git branching such as git flow. Presently, we have two active branches in WebAnno: master and 3.3.x. 3.3.x should get all the bug fixes, so I commit them there and then merge the whole 3.3.x branch into master afterwards. New features usually target the master branch - they are not merged back into the 3.3.x branch (although I might target the 3.3.x branch when doing a PR - retargeting a PR from 3.3.x to master is easier than from master to 3.3.x. For refactorings, it kind of depends... they should target the master branch, but if they are trivial, I might target the 3.3.x branch to facilitate merges later on.

kazukisan:
Thanks for the info! And as long as it's clear how the branches are used, I think that's fine.

It was just that to me, 3.3.x feels like the typical master branch with clean commits, build, and deployment, whereas the current master branch in WebAnno feels more like feature / experiment branches to be later pulled into master once finalized and stable. But it's just how I am used to, and not sure what I am used to is the best practice.

So, there is no need to change branching strategies, and thanks for again for WebAnno work!

PR #810, opened 2018-01-08
Fix #809 Unnecessary CPU load caused by autologout timer

### LINKED ISSUE #809: Webanno webpages eat CPU when idling
When a Webanno webpage is idling in an active browser tab, it causes constant CPU load, which it probably shouldn't. It looks like the autologout timer is called every second, updating the timer label causes cascading redraw.

Rentier:
Reducing the refresh rate reduces the load on my machine from constant 10% to 0% with spikes of 10.

### DIFF
--- a/webanno-ui-core/src/main/java/de/tudarmstadt/ukp/clarin/webanno/ui/core/logout/LogoutPanel.html
+++ b/webanno-ui-core/src/main/java/de/tudarmstadt/ukp/clarin/webanno/ui/core/logout/LogoutPanel.html
@@ -42,11 +42,12 @@
           };
               
           this.start = function() {
+               const TICK_RATE = 10 * 1000; // 10 seconds
                if (this.timer) {
                    clearInterval(instance.timer);
                }
                started = new Date();
-               timer = setInterval(instance.tick, 1000);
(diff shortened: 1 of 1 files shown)


reckart:
If there are only 60 seconds remaining, then the display switches from minutes to seconds. With the present patch, the user will not see how many seconds are remaining in "real time" as the update happens only every 10 seconds. One option would be to reduce the update time when switching to display-seconds mode. Another option would be to fix the re-rendering problem in a different way, e.g. by somehow jailing the timer in a fixed-size area which does not trigger refreshes in other parts of the view, primarily in the brat view.

Rentier:
Where is the code that switches from minutes to seconds? I would remove that feature for simplicity. Changing the tickrate means to reschedule the timer which is also more complicated. I do not know how to jail the timer, my CSS is not good enough and Google did not help.

reckart:
Jenkins, test this please.

reckart:
@Rentier Could you please verify whether the approach I have committed here also resolves the issue for you?

Rentier:
@reckart The unnecessary CPU consumption and redraw is gone. Your fix is much better than mine.

Issue #813, opened 2018-01-09
Generate additional internal events on user actions


reckart:
Generate additional internal events on user actions, e.g. when users create, update, or delete annotations.

PR #814, opened 2018-01-09
#813 - Generate additional internal events on user actions

reckart:
- Added SpanDeleted and SpanCreate events
- Added FeatureValueUpdated event


reckart:
Merged via CLI into master.

Issue #842, opened 2018-01-31
Extend forward annotation to a feature without a tagset


seyyaw:
Currently, forward annotation is possible only when there is a tagset on the feature. Add a possibility to enable forward annotation on a feature without a tagset too. This is specially useful for layers like lemma and Orthography correction

PR #843, opened 2018-02-02
#842 - Extend forward annotation to a feature without a tagset

seyyaw:
Extend forward annotation to free-text annotation feature when
1) it is locked-to-token annotation layer
2) There is only one annotation feature which is enabled and visible at a time. If there multiple features, they should be invisibile and disable to allow forward annotation for free-text annotation
How it works
1) enable forward annotation (using the checkbox)
2) type the annotation in the text-field and pres the ENTER KEY
If nothing is typed in the text-field, annotation will be delted and focus will move to the next token

### DIFF
--- a/webanno-api-annotation/src/main/java/de/tudarmstadt/ukp/clarin/webanno/api/annotation/action/AnnotationActionHandler.java
+++ b/webanno-api-annotation/src/main/java/de/tudarmstadt/ukp/clarin/webanno/api/annotation/action/AnnotationActionHandler.java
@@ -28,7 +28,14 @@ import de.tudarmstadt.ukp.clarin.webanno.api.annotation.model.VID;
 
 public interface AnnotationActionHandler
 {
+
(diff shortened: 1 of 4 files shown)


seyyaw:
Thanks @reckart . How do you change the base branch ?

reckart:
You can change the base by pressing on "edit" next to the title. It is a bit hidden ;)

I'll try to review the PR over the weekend.

reckart:
@seyyaw I have had a look at the code. If I see it correctly, you have separated the code paths for forward and non-forward annotation which were previously both handled in `actionCreateOrUpdate`. This seems to have introduced some duplicate code. I need more time to better understand this.

reckart:
Jenkins, could you please test this?

seyyaw:
@reckart I thought it is better to separate the logic even if there is large amount of code overlaps. The other option is to add more parameters to the method but then we should change method calls in several places. Let me know your thought on this.

reckart:

Issue #830, opened 2018-01-18
User may accidentally shut down WebAnno during startup


reckart:
When webanno is started (not headless), after some time it puts up a little
window with a "Shutdown" button.

On Windows, the box and the button are "selected".

This has an unfortunate consequence.  If while waiting for webanno to
initialize, you have another window open (like an email composition window) and
are typing, then what happens is that in the middle of typing some word, the
WebAnno dialog box/buttom appear and become selected, and the next keystroke
goes to that window / button, essentially "pushing" (by accident) the shutdown
button, which then shuts things down!

mbugert:
> On Windows, the box and the button are "selected".
I'm having the same issue on Ubuntu.

mbugert:
See pull request #848.

reckart:
I wonder actually why the window pops to the front at all.... does this happen with all new windows that are opened by a "background" application?

reckart:
Seems to be a thing of JDialog:  https://stackoverflow.com/questions/823795/jdialog-lets-main-application-lose-focus

reckart:
Actually, it might help to simply make the dialog **not** modal...

PR #848, opened 2018-02-12
#830 - User may accidentally shut down WebAnno during startup

mbugert:
Dialog now opens focusing the message instead of the shutdown button.

I'm not 100% happy with the fix though. Users can involuntarily shift the focus back to the shutdown button if they press tab in the wrong moment. Of course more advanced "missile switch cover" solutions could be built, but the underlying problem is rather that the dialog opens several seconds after opening webanno. Merging the splashscreen with the shutdown dialog would be the best solution UI-wise.


reckart:

reckart:
Found a better fix.

reckart:
Found a better solution.

Issue #879, opened 2018-03-05
Unable to delete automation project


kazukisan:
From 3.3.x

When I try to delete an automation project I am seeing:
```
Error: org.hibernate.exception.ConstraintViolationException: could not execute statement
```
in the project deletion dialog box.

```
2018-03-05 11:55:15 INFO [kazuki] DocumentServiceImpl - Removed source document [i like things 2.txt](1783) from project [automation](31)
2018-03-05 11:55:15 WARN [kazuki] SqlExceptionHelper - SQL Error: 1451, SQLState: 23000
2018-03-05 11:55:15 ERROR [kazuki] SqlExceptionHelper - Cannot delete or update a parent row: a foreign key constraint fails (`webanno`.`mira_template`, CONSTRAINT `FK_3rq9eim60vi4p3h8us1g93h9` FOREIGN KEY (`trainFeature`) REFERENCES `annotation_feature` (`id`))
2018-03-05 11:55:15 WARN [kazuki] SqlExceptionHelper - SQL Warning Code: 1451, SQLState: 23000
2018-03-05 11:55:15 WARN [kazuki] SqlExceptionHelper - Cannot delete or update a parent row: a foreign key constraint fails (`webanno`.`mira_template`, CONSTRAINT `FK_3rq9eim60vi4p3h8us1g93h9` FOREIGN KEY (`trainFeature`) REFERENCES `annotation_feature` (`id`))
2018-03-05 11:55:15 ERROR [kazuki] ProjectPage - Error: org.hibernate.exception.ConstraintViolationException: could not execute statement
javax.persistence.PersistenceException: org.hibernate.exception.ConstraintViolationException: could not execute statement
	at org.hibernate.jpa.spi.AbstractEntityManagerImpl.convert(AbstractEntityManagerImpl.java:1763) ~[hibernate-entitymanager-4.3.11.Final.jar:4.3.11.Final]
	at org.hibernate.jpa.spi.AbstractEntityManagerImpl.convert(AbstractEntityManagerImpl.java:1677) ~[hibernate-entitymanager-4.3.11.Final.jar:4.3.11.Final]
	at org.hibernate.jpa.internal.QueryImpl.getResultList(QueryImpl.java:458) ~[hibernate-entitymanager-4.3.11.Final.jar:4.3.11.Final]
[...]

PR #880, opened 2018-03-07
#879 Unable to delete automation project

seyyaw:
set TRAIN and OTHER features to null before deleting the Mira Template


reckart:

Issue #881, opened 2018-03-08
Order of icons in left annotation sidebar is not stable


Rentier:
The order in which the tabs of the left sidebar on the annotation page are shown changes from time to time. 

Solution: Sort by order if it is specified, use display name for breaking ties.

reckart:
PR merged.

PR #882, opened 2018-03-08
#881 Use stable order in left side bar

Rentier:
- Use order for sorting if existing, else use display name


reckart:

Rentier:
I looked into the source code of [OrderedComparator](https://github.com/spring-projects/spring-framework/blob/master/spring-core/src/main/java/org/springframework/core/OrderComparator.java). It does not check for the annotation `Order`. If you want that, then you would need to use [AnnotationAwareOrderComparator](https://github.com/spring-projects/spring-framework/blob/master/spring-core/src/main/java/org/springframework/core/annotation/AnnotationAwareOrderComparator.java).  I will fix it.

reckart:
@Rentier good point - I guess I need to make changes in other places as well then ;) Thanks!

Rentier:
I construct now my own comparator, as the types with Java 8 `thenComparing` do not work out.

reckart:
@Rentier can you resolve conflicts please so I can merge?

Issue #890, opened 2018-03-20
listAccessibleProjects should include projects in which the user is an annotator or curator


betoboullosa:
The api needs to differentiate between projects that are "accessible" and "manageable".

reckart:
Which issue/problem does this change address?

reckart:
Jenkins, can you test this please?

PR #891, opened 2018-03-20
#890 - listAccessibleProjects should include projects in which the user is an annotator or curator

betoboullosa:
- listAccessibleProjects now lists all projects in which the user is
admin / annotator / curator
- New method listManageableProjects to select only the projects in which
the user is a manager


reckart:

PR #915, opened 2018-04-18
Add Ajax Event in AnnotationDetailEditorPanel

matchlesswei:
- Add Ajax Event in AnnotationDetailEditorPanel, so that if user annotates in the right sidebar, active learning sidebar could receive this event and update recommendation in the left sidebar.
- Please review the event send location, event file name and event file location
- Please note that the current webanno master branch is 3.5.0, but inception master uses webanno branch 3.4.0


reckart:

Issue #930, opened 2018-05-04
Project initialization does not take into account the missing dependencies of a layer


betoboullosa:
The initializeProject method in the AnnotationSchemaService does not differentiate between two potential error conditions when initializing the layers of a project: 1) circular dependencies and 2) missing dependencies.

Currently, both conditions lead to the same exception, namely "Circular initializer dependencies". 

This is not a problem in the current default project initialization but might cause problems when someone tries to programmatically create a project (e. g., in a test unit).

PR #931, opened 2018-05-04
#930 - Project initialization does not take into account the missing dependencies of a layer

betoboullosa:
- initializeProject now tests for missing dependencies and generates the
proper exception message


reckart:

Issue #838, opened 2018-01-29
If "Forward annotation" is not allowed on a layer, hide the option


reckart:
If "Forward annotation" is not allowed on a layer, hide the option instead of just disabling it.

PR #993, opened 2018-06-14
#838 - If "Forward annotation" is not allowed on a layer, hide the option

reckart:
- Hide the form group of forwarding is not possible


reckart:

Issue #1000, opened 2018-06-19
"Cancel" button in user management has no effect


reckart:
Pressing the "cancel" button on the user management page has no effect.

PR #1008, opened 2018-06-27
#1000 - "Cancel" button in user management has no effect

reckart:
- Add user list and detail panel to the AJAX refresh


reckart:

Issue #892, opened 2018-03-20
Do not load the project page if the user does not have the proper rights


betoboullosa:
Currently, a user can directly access the project page by its URL, even though he does not have permission to access it, i.e., even though he does not see the link to it on the homepage. The page should not be loaded in that case - an "access denied" message would be shown instead.

reckart:
Wicket has a built-in access control mechanism. Need to see if that can be used here or if we need a custom measure.

PR #1014, opened 2018-07-08
#892 - Do not load the project page if the user does not have the proper rights

reckart:
- Added a custom authorization strategy that takes into account the menu item visibility


reckart:

PR #1028, opened 2018-07-30
fixing context path typo

shouro:


reckart:
@shouro Actually, both [works](https://docs.spring.io/spring-boot/docs/current/reference/html/boot-features-external-config.html#boot-features-external-config-relaxed-binding). Did you run into problems using the CamelCase variant?

shouro:
@reckart  I got it now. I missed that part.

PR #1031, opened 2018-08-01
Add check for multiple ingoing edges

thomaskrause:
We have a corpus that was annotated with an older version of WebAnno, where the annotators could accidentally add dependency relations with the same target token. This will fail to export with both the CONLL-2009 and the WebAnnoTSV exporter and we now have to correct all the files manually, since there is no way to output these to any format we can read.

Since which edge should be removed (or reversed) only can decided manually, I added a check to the CAS Doctor that finds all offending instances and outputs a message that includes the sentence number (useful to jump directly to the problem) and the actual offending dependency edges.

checkoutputexample

I'm fairly new to the data model, if there is anything I should improve in the implementation please let me know.

### DIFF
--- a/webanno-diag/pom.xml
+++ b/webanno-diag/pom.xml
@@ -92,7 +92,12 @@
     </dependency>
     <dependency>
       <groupId>org.springframework.boot</groupId>
-      <artifactId>spring-boot-starter-test</artifactId>
+      <artifactId>spring-boot-test</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.mockito</groupId>
+      <artifactId>mockito-core</artifactId>
       <scope>test</scope>
     </dependency>
     <dependency>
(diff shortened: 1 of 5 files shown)


reckart:
@thomaskrause Unlike the CoNLL formats, the WebAnno TSV 3 format should be able to handle arbitrary relations - multiple incoming edges should not pose a problem. Could you open a separate issue for that problem and provide some instructions on how to reproduce the issue?

reckart:

reckart:
@thomaskrause The PR does not pass the build. In particular, checkstyle reports various problems. If your IDE does not support, please run a Maven build on the command line in order to see checkstyle's messages.

reckart:

reckart:
@thomaskrause Build looks good, thanks! 

Could you please also include a section in the CAS Doctor documentation (`webanno-diag/src/main/resources/META-INF/asciidoc/developer-guide/casdoctor.adoc) explaining your new check? When you build WebAnno on the command line, you can find the generated documentation under `webanno-webapp/target/generated
-docs`).

A unit test would also be very much appreciated (cf. `webanno-diag/src/test/java/de/tudarmstadt/ukp/clarin/webanno/diag/checks/AllAnnotationsIndexedCheckTest.java`).

thomaskrause:
I added the documentation and some tests. I needed to add a dependency to `spring-boot-starter-test` to allow mocking the `AnnotationSchemaService` bean easily and this is ok.

reckart:

reckart:

thomaskrause:

reckart:

@thomaskrause I'm afraid Jenkins won't listen to you unless you are whitelisted.

thomaskrause:
> @thomaskrause I'm afraid Jenkins won't listen to you unless you are whitelisted.

I almost suspected this, but I just had to try :relaxed:

reckart:

PR #1125, opened 2018-11-22
Feature/1124 - Generic render method for annotation editors

wpmp:
- move render method from BratAnnotationEditor to AnnotationEditorBase
- move getLayersToRender method from BratAnnotationEditor to AnnotationEditorBase


reckart:

Issue #1138, opened 2018-11-29
Combobox limit 25


sarschu:
I am using String features with large tagsets (up to 5000 tags) which I navigate by entering a couple of characters into the box to prefilter them before I scoll through them to pick the one I am looking for. Unfortunately, the number of options showing in this dropdown seems to be limited to 25. Is there a way to increase this limit?

reckart:
The limit is currently hard-coded, but we can make it configurable on a global base via a setting in the configuration file. 

It could also be made configurable through the UI, but not in the short term because I would rather do this along with a complete refactoring of the per-user preference management code.

sarschu:
A configuration parameter in the configuration file sounds like a good solution to me.

reckart:
@sarschu also, it might be sensible to increase the default limit - what would you think would be a sensible default limit? 100? 250?

sarschu:
I guess 100 should be enough. Usually you can filter far enough by entering enough characters to have not more than 100 feature values containing the substring

PR #1139, opened 2018-11-29
#1138 - Combobox limit 25

reckart:
- Increased default limit for auto-complete field to 100
- Made activation threshold and limit for auto-complete field configurable via settings.properties


reckart:

Issue #1197, opened 2019-01-23
Codebook Annotation support in WebAnno


seyyaw:
codebooks are document-level annotations. The project page should allow creating codebooks. Codebooks will have a category and each category can have a tag set. The annotation page should allow selecting a tag for each category of a codebook annotation.

reckart:
@seyyaw could you please rename the branch to `feature/1197-Codebook-Annotation-support`

reckart:
Please also stick to the commit message convention:

```
#[ISSUE NUMBER] - [ISSUE TITLE]
[EMPTY LINE]
- [CHANGE 1]
- [CHANGE 2]
- [...]
```

PR #1202, opened 2019-01-24
#1197 - Codebook annotation support

reckart:
**What's in the PR**
* Codebook annotation: Codebook annotation is a document level annotation. It has a different annotation editor than the existing WebAnno layer based annotation. Codebook categories should be created in the project settings. For each Category, it will have a tagset with multiple tags to annotate.
* Codebook Curation: Curation of codebook annotation similar to the exiting curation functionality but it is document level curation
* Import/Export document support: Allows to import and export codebook annotations. From the annotation and curation page, exporting is possible only for the current document. From the project settings page, export is aggregated per document and per annotator. For importing the annotation, the codebook category and the user should exist in the project. Furthermore, it allows importing excel documents, only from the first sheet, and has two columns. The first column is the document name, the second column is the content of the document. Hence the excel import allows importing multiple documents from a single spreadsheet.
* Import/Export codebook categories: Allow to export and re-import the codebook category settings.
* Codebook agreement: Similar to the existing agreement functionality but shows agreements for two users per a category

**How to test manually**
* Codebook category creation
   * Start WebAnno. It can be started with old WebAnno databses but the database structure will be changed then
   * Go to the project settings page and click **Coodebooks** tab
   * Press the **Create Categor** button. Give it a name for the category.
   * It is possible to create tags for each category
* Codebook Annotation: 
   * Open the annotation page
   * Go to the **Settings** page and select the **Show coodbook editor** checkbox. 
   * The codebook annotation editor will be displayed for the annotation task
* Curation Annotation
   * Open the curation interface and repeat the steps done for the codebook annotation page
   * The curation interface will show the curation panel and the suggestion panels. If there are disagreements, cick the **MERGE** button in the suggestion view to curate the annotation

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**


reckart:
@seyyaw right now, the changes affect a large number of modules. To assess the true impact of this PR on the architecture, could you please create a module `webanno-codebook` and move as much as possible of the new code there (this does not only include new classes/UI but also additions to existing classes which are not strictly required to be there, e.g. additions to `WebAnnoConst`, utility classes, services, etc.

seyyaw:
> @seyyaw right now, the changes affect a large number of modules. To assess the true impact of this PR on the architecture, could you please create a module `webanno-codebook` and move as much as possible of the new code there (this does not only include new classes/UI but also additions to existing classes which are not strictly required to be there, e.g. additions to `WebAnnoConst`, utility classes, services, etc. - New code which is already in new modules (like the Excel stuff) can remain there.

I have moved the codebook related codes to a separate module called `webanno-codebook`. The tags and tagsets are also shared among codebook and layer based annotation, hence moved to a separate `webanno-tagsets` module

reckart:
Thanks. That's a first step.

Why did you introduce a new `CodebookFeature` (along with supports, registry, etc.) instead of relying on the existing extensible feature mechanism?

reckart:
I can see that there are still many codebook-specific changes and additions in the `webanno-api-annotation` module, the `webanno-api-dao` module and the `webanno-model` module. The idea is that any changes to the core WebAnno modules are kept minimal and if there are changes, they should contribute to existing features and/or the overall modularization of the architecture. 

Can you please pull any code-book specific code out of existing classes and interfaces (in particular `AnnotationSchemaService`, `DocumentService`, `WebAnnoConst`) and move codebook specific classes/interfaces to the `webanno-codebook` module (in particular `CodebookSchemaService`, `CodebookExporter`, `CodebookConfigurationChangedEvent`, `Codebook`, `CodebookFeature`, `UimaStringCodebookTraitsEditor`, ``) as well as the codebook-specific changes to the `db-changelog.xml` files. As part of moving out the `db-changelog.xml`,  could you please merge master and resolve the conflict.

seyyaw:
Almost all codebook related codes moved to webanno-codebook and webanno-api-codebook modules.

seyyaw:
@reckart Do you have time to comment on the latest changes?

reckart:
@seyyaw I think it is now at a point where I actually need to try it out and spend a bit more time with it before I can give informed feedback, and I didn't get to that yet. But anyway, here is what I gleaned so far:

Actually, I have had a first look at the changes shortly after you have made them and I am pretty sure that there is still more codebook-specific stuff that should be moved into the codebook module, e.g. the methods you added to the `DocumentRepository`, `ColoringStrategy`. There seems to me to be no need to modify the code of the existing exporters, new codebook-specific exporters should be implemented inside the codebook module. I believe you can handle the cleanup simply by adding a cascade-on-delete at the database level and don't need to hook into schema change events and thus your changes to `AnnotationSchemaServiceEventAdapter` (if you still want to hook into that event, you should create a new adapter in the codebook module and handle the event there).

The goal is that there are no changes to the code outside the codebook module unless they are in general necessary to improve the modularity of the system and are required to allow the codebook module to *hook in*. But none of the other modules should actually depend on the codebook module or on any entities it provides in any way.

It would be great if you could pull in the latest changes from master and resolve the conflict!

seyyaw:
Thanks @reckart.  I will let you know once I have resolved and move the remaining codebook related stuff out of the other modules.

seyyaw:
@reckart Most of the dependencies are now cleaned up. It would be nice you go through the changes. In the `AnnotationSchemaServiceImpl` class, I still use codebook service to get the codebook types to build the UIMA types for codebook annotation. Is that ok or tell me how to do it without hooking it up in this class.

reckart:
@seyyaw I'll have a look at it (might take a while). Meanwhile, could you please look at why the build for this PR is failing and fix that?

seyyaw:
@reckart it compiles from the command line. I saw the Jenkins output and it just says one error but I could not find the error. Where I could I get that specific build failure?

reckart:
```
[ERROR] E:\svc_jenkins\workspace\WebAnno (GitHub) (PR builder)\webanno-ui-project\src\main\java\de\tudarmstadt\ukp\clarin\webanno\ui\project\ProjectPage.java:76: error: reference not found
[ERROR] * and name of the Project The {@link ProjectTagSetsPanel} is used to add {@link TagSet} and
[ERROR] ^
[ERROR] 
```

reckart:
I roughly went through the changes and addressed several things as I was going along. It seems there are still a number of changes that need to be done to be base architecture in order to accommodate this extension. But first there are a couple of smaller things that can be fixed before that:

- [ ] Move codebook I18N strings out of WicketApplicationBase.properties and into an appropriate place in the codebook module
- [ ] Move new images from webanno-ui-core to codebook module
- [ ] Remove codebook-related changes from the remote API module. It seems to me that they are not really necessary there.
- [ ] Move codebook CSS out of BaseLayout.css and custom.less and into an appropriate CSS file in the codebook module
- [ ] Move ExportedCodebook and CodebookFeature to codebook module and drop codebook reference from ExportedProject. Instead of explicitly adding the `codebooks` field to ExportedProject, have the exporter/importer for codebooks use the `properties` field (cf. `AutomationMiraTemplateExporter.importData/exportData` calling `aExProject.setProperty(...)` and `aExProject.getArrayProperty(...)`).

I think what comes after that will be more effort and more fundamental and I need more time to investigate here.

seyyaw:
@reckart smaller issues are now fixed.

seyyaw:
@reckart It would be nice to finish this PR as early as possible. Keeping it up to the master is getting challenging. At least if the main architecture is merged, we can keep on working on additional changes over time.

reckart:
There are various things in this PR which are still affecting the core code. These need to be resolved so this plugin can be maintained independently from WebAnno. This is from my latest set of notes:

1. Annotation UI
    - Turn codebook sidebar into a proper sidebar and remove it from AnnotationPage
    - Preferences for codebooks need to be moved out of the AnnotationPreference dialog
2. Document-level annotations
    - AnnotationSchemaServiceImpl must not depend on codebook
    - Remove special handling of "webanno.codebook" type prefix from Tsv3XCasDocumentBuilder
    - Remove special codebook stuff from TagSetSelectionPanel
3. Curation
    - Remove codebookService from SuggestionBuilder
    - Need to read change to curation panel in more detail
4. Import/export
    - Remove special code for codebook from ImportDocumentsPanel
    - Remove special handling of "isDocumentLevel" from AnnotationDocumentExporter
    - Remove special handling of codebook from ExportDocumentDialogContent (isDocumentLevel)
    - Remove special handling of WebAnnoCsvFormatSupport from CuratedDocumentsExporter

Wrt 1. Annotation UI: I see three ways to handle this
- create a generic support for multiple editors being visible at the same time (e.g. brat + codebook); however, I don't see this happen very soon
- create a separate codebook annotation page and remove your changes from the annotation page; this is very ugly and has quite a bit of maintenance overhead
- *realistic solution*: turn the codebook UI into a proper sidebar tab. It is visible when the user opens this sidebar. However, the main brat editor cannot be disabled. The settings from the preferences can be removed.

Wrt 2. Document-level annotations:
- you could look at my implementation of Document-level annotations in INCEpTION to see how it is possible to add a new layer type without having to modify core code.

Wrt 3. Curation
- curation needs to be modularized to support plugging in new layer/feature types; is there actually an UI for curating codebook annotations? This could be dropped (i.e. no curation for codebook) and done later.

Wrt 4. Import/export
- My understanding is that codebook annotations are exported as one file for an entire project, i.e. some kind of table where each row is a document and the codebook labels are in the columns. Is that correct? If this is the case, it requires changes to the core code to make the import/export mechanism more powerful and would need to be discussed/implemented in a separate issue/PRs.

I am working on a mechanism that allows the development of independent plugins. That means the user can first install WebAnno and afterwards download and install a plugin. This also means that the plugin code doesn't have to be in the WebAnno repo anymore and that it can be developed and maintained separately. This is important for me to ensure that the maintenance overhead of the core code remains controllable. So once the issues above are resolved, we should move the code from the PR (which then does not contain any core code changes anymore) to a new repo specifically for this plugin.

reckart:
Btw. a basic plugin mechanism is now available: https://github.com/webanno/webanno/issues/1368

... and we're also inching towards making first use of it in INCEpTION by turning some experimental and example modules from built-in modules into plugins: https://github.com/inception-project/inception/pull/1244

Issue #1296, opened 2019-05-10
Publish Relation Created Event


UWinch:
**Is your feature request related to a problem? Please describe.**
I would like to publish an event when Relations are created *or deleted*. I will use this to identify annotation events for resuming previous work in Inception. 

**Describe the solution you'd like**
Modify `RelationAdapter` to publish a new event when relations are created. Implement a new event type for this.

**Additional context**
For development of https://github.com/inception-project/inception/issues/804

Update 05-14-2019: The event should also be used to detect deletions of relations.

PR #1298, opened 2019-05-14
1296 relation update event

UWinch:
**What's in the PR**
* closes https://github.com/webanno/webanno/issues/1296
* I would like to publish an event when Relations are created or deleted. I will use this to identify annotation events for resuming previous work in Inception.

**How to test manually**
* checkout inception PR https://github.com/inception-project/inception/pull/1142
* this PR will show last activities as events on the project dashboard page
* try annotating documents with dependency relations (e.g. POS) 
* try deleting dependencies
* when opening the dashboard page, the last event should be shown

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**

### DIFF
--- a/webanno-api-annotation/src/main/java/de/tudarmstadt/ukp/clarin/webanno/api/annotation/event/RelationUpdateEvent.java
+++ b/webanno-api-annotation/src/main/java/de/tudarmstadt/ukp/clarin/webanno/api/annotation/event/RelationUpdateEvent.java
@@ -24,7 +24,7 @@ import org.springframework.context.ApplicationEvent;
 import de.tudarmstadt.ukp.clarin.webanno.api.annotation.adapter.RelationAdapter;
 import de.tudarmstadt.ukp.clarin.webanno.model.SourceDocument;
 
-public class RelationCreatedEvent extends ApplicationEvent
+public class RelationUpdateEvent extends ApplicationEvent
 {
     private static final long serialVersionUID = -8621413642390759892L;
     
@@ -32,7 +32,7 @@ public class RelationCreatedEvent extends ApplicationEvent
     private final String user;
     private final AnnotationFS annotation;
 
-    public RelationCreatedEvent(Object aSource, SourceDocument aDocument, String aUser,
+    public RelationUpdateEvent(Object aSource, SourceDocument aDocument, String aUser,
             AnnotationFS aAnnotation)
     {
         super(aSource);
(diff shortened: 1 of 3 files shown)


UWinch:
Can be reviewed and merged

webanno-api-annotation/src/main/java/de/tudarmstadt/ukp/clarin/webanno/api/annotation/event/RelationUpdateEvent.java
```diff
@@ -0,0 +1,91 @@
...
+import org.apache.uima.cas.Feature;
+import org.apache.uima.cas.text.AnnotationFS;
+import org.springframework.context.ApplicationEvent;
+
+import de.tudarmstadt.ukp.clarin.webanno.api.annotation.adapter.RelationAdapter;
+import de.tudarmstadt.ukp.clarin.webanno.model.SourceDocument;
+
+public class RelationUpdateEvent extends ApplicationEvent
+{
+    private static final long serialVersionUID = -8621413642390759892L;
```
reckart: I would suggest to implement HybridApplicationUIEvent such that the UI can listen to this event a well.

webanno-api-annotation/src/main/java/de/tudarmstadt/ukp/clarin/webanno/api/annotation/event/RelationUpdateEvent.java
```diff
@@ -0,0 +1,91 @@
...
+
+    public AnnotationFS getAnnotation()
+    {
+        return annotation;
+    }
+    
+    @Override
+    public String toString()
+    {
+        String govFeatureName = ((RelationAdapter) getSource()).getSourceFeatureName();
```
reckart: I think we should avoid casting the source to RelationAdapter here because there is no guarantee that the source actually is a RelationAdapter (cf. source being an Object in the constructor). An option might be to explicitly pass the relation adapter as an additional type-checked argument. I'd also suggest considering not to pass the entire adapter around but extract from it what information is relevant. Mind that the `toString()` method should only have meaning for debugging and in other cases shouldn't really be used. Yet another option might be to check if the source is actually an instance of RelationAdapter and if yes use it for the toString method... but it would seem a bit hacky because in some situation toString() would produce one format and in others another format.

reckart:

webanno-api-annotation/src/main/java/de/tudarmstadt/ukp/clarin/webanno/api/annotation/event/RelationUpdateEvent.java
UWinch: Thanks for the suggestion. I implemented the requested changes by also passing the source annotation to the vent.

reckart:
I'm planning to merge this after the next beta.

Issue #1390, opened 2019-08-11
Move focus to feature name field when creating new feature


reckart:
**Is your feature request related to a problem? Please describe.**
When creating a new feature in the layer settings, I need to manually click on the feature name field to enter a name.

**Describe the solution you'd like**
When creating a new feature, the input focus should move directly to the name field for the new feature.

PR #1391, opened 2019-08-11
#1390 - Move focus to feature name field when creating new feature

reckart:
**What's in the PR**
- Use AJAX to load the feature detail panel
- Try putting the focus on the name field (which should work but for some reason has no effect so far...)

**How to test manually**
* Create a new feature
* See if the focus goes to the feature name input field

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**


reckart:
Should work - even in the Wicket AJAX debug log in the browser, it says that the focus is moved, but the focus does not appear in the field... I even tried to manually focus the name input field using JS commands in the browser console but without success... no idea...

Issue #1370, opened 2019-07-22
Extend VID for use in editor-extensions


UWinch:
**Is your feature request related to a problem? Please describe.**
I would like to extend the VID with extension-specific contents and have the extension be automatically set during the rendering process. I hope this will allow me to add several annotations for rendering in an editor extension without re-implementing all rendering code that currently creates VIDs.

**Describe alternatives you've considered**
I thought about re-implementing part of the rendering process similar to the RecommenderEditorExtension and the RecommendationRenderer. However, this requires also implementing element-specific renderers like the SpanRenderer (cf. RecommendationSpanRenderer).

**Additional context**
Will use this for this PR https://github.com/inception-project/inception/issues/1256

reckart:
The VID is already pretty heavy as it is. What do you want to add to it?

Why can't we simply use the existing pre-renderer and then post-process it's output?

UWinch:
I would like to distinguish between annotations of different users when I add all of them for curation. My understanding is that if these annotations are at the same place, they will not be rendered because they have the same VID. It also looks to me that the extensionID which can already be added via VID constructor is only set for VObjects that are explicitly created by the extension, so if I just want to use the existing pre-rendering process, this is not done.

I also thought about post-processing the VDocument after pre-rendering, but I could only come up with changing all VIDs for all elements that are added during one pre-rendering run.
What were your ideas for this?

reckart:
I understand the need to improve the VID. I would suggest to redesign it in such a way that when it contains an extension ID, the format of the rest is actually not defined, i.e. an extension has to parse out the contents itself, potentially even by parsing it again as a VID (e.g. `VID.parse(VID.parse("ext:1@POS`).getExtensionPayload()) == VID.parse("1@POS`)`. That would allow an extension to pack in any additional information into the payload. Does that make sense to you?

So you want to extend the `PreRenderer.render()` method such that you can pass an optional extension ID or even some kind of `VIDFactory` in?

> I also thought about post-processing the VDocument after pre-rendering, but I could only come up with changing all VIDs for all elements that are added during one pre-rendering run.
> What were your ideas for this?

I guess the approach would involve pre-rendering into an empty VDocument for every user and then copying the information into the main VDocument - and during that process also adjust VIDs.

UWinch:
[...]

PR #1396, opened 2019-08-13
#1370  Support curation from annotationpage in INCEpTION

UWinch:
Changes needed to integrate curation into annotation page in inception. Support for https://github.com/inception-project/inception/pull/1257

**What's in the PR**
* Extend VID for use in editor-extensions: VID format can either be previous format without optional extension-ID or contain extensionID, annotationID and extension payload. This payload is defined by the specific extension and parsed by it.
* Return result with addresses of updated/inserted annotationfs from casmerge
* Make sidebars optional
  * Re-factor annotationpage to init params before component creation -> can now check for curator status before configuring curation-sidebar
  * default implemented method in sidebars that check if sidebar should be added to page

* Closes https://github.com/webanno/webanno/issues/1370
* Closes https://github.com/webanno/webanno/issues/1401

**How to test manually**
* Test annotating, check that everything works as expected.
* Test curation, check that everything works as expected.

**Automatic testing**
* [x] PR updates unit tests


webanno-brat/src/main/java/de/tudarmstadt/ukp/clarin/webanno/brat/annotation/BratAnnotationEditor.java
```diff
@@ -172,6 +172,9 @@ else if (!request.getParameterValue(PARAM_ID).isEmpty()) {
                 else {
                     paramId = VID.parseOptional(request.getParameterValue(PARAM_ARC_ID).toString());
                 }
+                if (paramId.isSynthetic()) {
```
reckart: Can this not happen in the "isSynthetic" branch below? I believe that here we only need to check if the VID contains an extension ID so we can route events to the extension. The receiving extension then needs to see what it can get from the payload of the VID.

UWinch:
Can be reviewed for https://github.com/inception-project/inception/pull/1257. Please test extensively because I changed the `AnnotationPage` (mostly to allow sidebars to become optional).

webanno-api-annotation/src/main/java/de/tudarmstadt/ukp/clarin/webanno/api/annotation/rendering/model/VDocument.java
```diff
@@ -137,4 +137,14 @@ public VObject get(VID aVid)
...
     }
+
+    public void add(VObject aVobj)
+    {
+        if (aVobj instanceof VSpan) {
+            add((VSpan) aVobj);
+        }
+        else if (aVobj instanceof VArc) {
+            add((VArc) aVobj);
+        }
```
reckart: This method should have an additional `else` block where an `IllegalArgumentException` is thrown when the object is neither a span nor an arc.

webanno-brat/src/main/java/de/tudarmstadt/ukp/clarin/webanno/brat/annotation/BratAnnotationEditor.java
```diff
@@ -310,6 +308,22 @@ else if (GetDocumentResponse.is(action)) {
...
         add(requestHandler);
     }
+    
+    private void fireExtensionAction(AjaxRequestTarget aTarget,
+            final IRequestParameters request, String action, VID paramId, CAS cas)
+        throws IOException, AnnotationException
+    {
+        Optional<Offsets> optOffsets = getOffsetsFromRequest(request, cas,
+                paramId);
+        if (optOffsets.isPresent()) {
```
reckart: Since the offsets are no longer used in `fireAction`, this can be simplified.

webanno-brat/src/main/java/de/tudarmstadt/ukp/clarin/webanno/brat/annotation/BratAnnotationEditor.java
```diff
@@ -398,8 +412,14 @@ private SpanAnnotationResponse actionSpan(AjaxRequestTarget aTarget, IRequestPar
     {
         // This is the span the user has marked in the browser in order to create a new slot-filler
         // annotation OR the span of an existing annotation which the user has selected.
-        Offsets userSelectedSpan = getOffsetsFromRequest(request, aCas, aSelectedAnnotation);
+        Optional<Offsets> optUserSelectedSpan = getOffsetsFromRequest(request, aCas,
```
reckart: How can the code arrive here if no offsets are present?

webanno-ui-annotation/src/main/java/de/tudarmstadt/ukp/clarin/webanno/ui/annotation/detail/AnnotationDetailEditorPanel.java
```diff
@@ -479,7 +479,8 @@ public void actionFillSlot(AjaxRequestTarget aTarget, CAS aCas, int aSlotFillerB
 
         AnnotatorState state = getModelObject();
         
-        if (isUserViewingOthersWork()) {
+        if (AnnotationPageBase.isUserViewingOthersWork(getModelObject(),
```
reckart: You could consider moving `isViewingOtherUsersWork` into `AnnotatorState` with only the current user as the argument. I think that would be nice than having it as a static method.

webanno-brat/src/main/java/de/tudarmstadt/ukp/clarin/webanno/brat/annotation/BratAnnotationEditor.java
UWinch: Hmm, there was probably a reason for this once upon a time, but I tested it just now and there seems to be no need for this check. You are right.

Issue #1404, opened 2019-08-20
Modify slot behavior for Annotations other than spans


wpmp:
**Is your feature request related to a problem? Please describe.**
Currently the slot behavior for link features only works when used within the annotation detail editor (right sidebar).
However in INCEpTION we want to create links between document and span annotations in another panel.
There is no selection made when trying to fill the slot.
The selection should be made if selection VID differs from the VID of the owner of the slot.

**Describe the solution you'd like**
Add the armed VID to the AnnotatorState.
Add checks if selected VID and armed VID differ, if so make a selection.

PR #1405, opened 2019-08-20
#1404 - Modify slot behavior for Annotations other than spans

wpmp:
**What's in the PR**
- Introduce new event interfaces to allow generically listening to all annotation creation or deletion evens
- allow ADEP.actionFillSlot() to be successful even if the armed slot is NOT in the ADEP and even if the link host is NOT an AnnotationFS
- Pass the entire feature state along when arming a slot to allow direct access to it when the slot is actually filled later on
- new RenderSlotsEvent is fired by AnnotatorStateImpl when setArmedSlot() is called

**How to test manually**
Do the following steps in INCEpTION:
1. Create a new document annotation layer with a link feature to a span annotation layer.
2. Open a document
3. Create a document annotation with the type of your new layer
4. Select the new slot.
5. Annotate something in the text.
6. Link should be created

Also try out having a span layer with a link to another span.
Then have a document annotation and a span annotation both with links open and play around to see if the slots are armed/disarmed correctly.

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**


reckart:

webanno-api-annotation/src/main/java/de/tudarmstadt/ukp/clarin/webanno/api/annotation/model/AnnotatorStateImpl.java
```diff
@@ -163,10 +165,17 @@
...
     private List<Unit> visibleUnits;
     
+    private transient ApplicationEventPublisher eventPublisher;
+    
     public AnnotatorStateImpl(Mode aMode)
     {
         mode = aMode;
     }
+    
+    public AnnotatorStateImpl(Mode aMode, ApplicationEventPublisher aEventPublisher) {
```
reckart: You can't keep the ApplicationEventPublisher in the state - it is not serializable.

wpmp: I made it transient.
Should I avoid doing this?

reckart: Yes. The state object is a data object. I think it shouldn't go around sending events. Also, I think we don't need to send a Spring event via the ApplicationEventPublisher - a Wicket event should suffice. Cf. check e.g. how `RenderAnnotationsEvent` is sent and recieved (via `@OnEvent`).

reckart: With Wicket events, you could even send the event, cf. how `HybridApplicationUIEventRouter` fetches the current Page (if one is available) and then uses it to send an event.

wpmp: Alright, thanks I'll have a look.
But from where should these events be sent then?
From the editors (`LinkFeatureEditor`, `SubjectObjectFeatureEditor`, etc.)?
I added this to the `AnnotatorState` because it is the place where a slot is armed and it would ensure that a `SlotArmedEvent` is always sent.
If we do it from the editors we'd need to add it everywhere where `setArmedSlot` ist called, right?

wpmp: I changed it now for the `FeatureLinkEditor`.
Is this the way it should work?

reckart: I edited my comment a couple of times. I think it should be ok to embed your `sendRenderSlotsEvent()` methods into `setArmedSlot()` - at least that's better than having to fire the events manually every time `setArmedSlot()` is called.

wpmp:
I think this PR is ready for a review

reckart:

reckart:

Issue #1420, opened 2019-09-03
Hide layer in Annotation UI if only one layer is available


wpmp:
**Is your feature request related to a problem? Please describe.**
If only one layer is enabled in a project the annotation sidebar should not show the layer selection ~~and the layer of a selected annotation~~.

**Describe the solution you'd like**
Layer selection ~~and layer of selected annotation~~ is hidden.

PR #1421, opened 2019-09-03
#1420 - Hide layer in Annotation UI if only one layer is available

wpmp:
**What's in the PR**
- hide layer selection and annotation layer from AnnotationDetailEditorPanel if only one span layer is available in a project

**How to test manually**
1. Make a new project
2. Have only one span layer
3. Open the new project
4. Layer selection should be not visible
5. Create an annotation
6. Layer of annotation should be not visible
7. Open old project with multiple span layers
8. Layer selection and layer of annotation should still be visible

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**

### DIFF
--- a/webanno-api-annotation/src/main/java/de/tudarmstadt/ukp/clarin/webanno/api/annotation/exception/IllegalFeatureValueException.java
+++ b/webanno-api-annotation/src/main/java/de/tudarmstadt/ukp/clarin/webanno/api/annotation/exception/IllegalFeatureValueException.java
@@ -1,0 +1,44 @@
+/*
+ * Copyright 2019
+ * Ubiquitous Knowledge Processing (UKP) Lab and FG Language Technology
+ * Technische Universität Darmstadt
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package de.tudarmstadt.ukp.clarin.webanno.api.annotation.exception;
+
+public class IllegalFeatureValueException
+    extends AnnotationException
+{
+    private static final long serialVersionUID = -5120144780593360107L;
+
+    public IllegalFeatureValueException()
+    {
+        super();
+    }
+    
+    public IllegalFeatureValueException(String message)
+    {
+        super(message);
+    }
+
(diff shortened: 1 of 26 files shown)


wpmp:
@reckart is `wicket:enclosure="selectedAnnotationLayer"` necessary?
To be honest I didn't really understand the purpose of it.
However it seems that stuff is also working without it?

reckart:
The `wicket:enclosure` wraps the label and the form input to which the label applies. What it does is: if the form input is set to being invisible, then the label (basically everything in the enclosure) is also made invisible. Otherwise, if you'd hide a field, its label and boilerplate HTML would still be hanging about.

reckart:
E.g. in your case, you'd make the `selectedAnnotationLayer` invisible and the surrounding enclosure would disappear as well - no need to introduce the `selectedAnnotationLayerContainer`.

wpmp:
Hm, I tried removing `selectedAnnotationLayerContainer` and adding the enclosure back again, however when I hide `selectedAnnotationLayer` nothing changes and the annotation layer field is still shown in the sidebar.
Do I have to do anything additional?

reckart:
I think you have not committed the change.

wpmp:
Oh, yes, sorry..
However I just thought about the following: what if I have just one span layer but also one or more relation layers?
I guess it is maybe better to only remove the layer selection but keep showing the layer of the currently selected annotation.
Otherwise you would not be able to see which relation layer is selected, right?

reckart:
True. The type of the selected layer should remain visible.

reckart:
Try using `selectedAnnotationLayer.add(LambdaBehavior.visibleWhen(...)` instead of explicitly calling setVisible.

wpmp:
Although we won't hide the selected layer now and we don't need this I tried it and it works.
Do you know why it does not work with calling `setVisible()` though?

reckart:
My guess: it's a timing issue. The information on which the decision is made may not be available at the time when you called `setVisible(...)` explicitly. The behavior kicks in every time the component is re-rendered and therefore always has the latest information to make its decision.

wpmp:
Hm, however the layer selection hides and its visibility is determined at the same point in code with the same condition, or do you mean something else?
But I guess using a `LambdaBehavior` in such a case is better anyway, isn't it?

reckart:
Same point in the code, but the action kicks in at a different time. You **add** the behavior in the same place in the code. But the behavior kicks in during the **onConfigure** part of the rendering life-cycle and at that point the lambda method that you passed to the behavior is called. The `LambdaBehavior.visibleWhen()` is a shortcut to creating an anonymous inner class for the component and overriding its `onConfigure` method.

wpmp:
Sorry I think I didn't express well what I meant:
Both, the layer selections as well as the selected layers visibility are determined in the same place.
One worked with calling `setVisibility()` the other didn't.
So this is what I am wondering about

reckart:
Hm, then it probably shouldn't make a difference...

wpmp:
Ok, should be ready for review now.

reckart:

Issue #1455, opened 2019-09-21
Larger string input features for annotations


wpmp:
**Is your feature request related to a problem? Please describe.**
Currently it is possible to add a string feature to an annotation layer.
However the string input size is fixed and only one line.
A possibility to modify the size of the string input like having multiple rows is not possible.

**Describe the solution you'd like**
An option to set the wanted number of rows for a string input field could be added.
Another possibility would be to open an overlay with a bigger string input field when clicking on the original input field

**Additional context**
inception-project/inception#1214

PR #1457, opened 2019-09-23
#1455 - Larger string input features for annotations

wpmp:
**What's in the PR**
- Add option for multiple rows and count of collapsed and expanded rows for primitive string features
- Use normal input textfield feature if multiple rows is disabled, otherwise use input textarea

**How to test manually**
* Create a new string feature for a layer and enable multiple rows
* Set collapsed and expanded rows
* See results in the annotation editor

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**

### DIFF
--- a/webanno-api-annotation/src/main/java/de/tudarmstadt/ukp/clarin/webanno/api/annotation/feature/PrimitiveUimaFeatureSupport.java
+++ b/webanno-api-annotation/src/main/java/de/tudarmstadt/ukp/clarin/webanno/api/annotation/feature/PrimitiveUimaFeatureSupport.java
@@ -20,6 +20,7 @@ package de.tudarmstadt.ukp.clarin.webanno.api.annotation.feature;
 
 import static java.util.Arrays.asList;
 
+import java.io.IOException;
 import java.util.Collections;
 import java.util.List;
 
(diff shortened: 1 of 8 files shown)


wpmp:
@reckart can you please have a look on this and tell me how you like the textarea?
There are now settings to have multiple rows for a string feature and if enabled you can set the number of rows for collapsed and expanded state:
textarea_settings
If multiple rows are enabled the string feature will render as a textarea, otherwise it will behave as it was before (simple textfield if no tagset is set, etc.)


The textarea still pushes down other features when expanded.
Do you think we should change this?


Also as you can see below it is possible to manually resize the textarea.
After you enter something into the textarea and lose focus it will change back to collapsed row size.
If you don't enter anything it will stay the size.
textarea_behavior


Another thing: If you press enter the cursor moves to a new line but then this will also trigger setting the feature and set the focus to the next feature.
What should we do about this?
We maybe could do something like shift+enter for new line like it is the case in some messengers.

reckart:

reckart:

reckart:

reckart:

reckart:

wpmp:
This one does not have any user documentation yet, because I wasn't sure if you like the changes as they are.
I could add some documentation later today?

reckart:
Good idea :) Drop me a not then when I can merge it.

wpmp:
@reckart  I added documentation, ready for review/merging

reckart:

Issue #1460, opened 2019-09-24
Boundary for number features


wpmp:
**Is your feature request related to a problem? Please describe.**
Currently, number features (integer/float) can be used for annotations.
However it is not possible to set upper and/or lower boundaries for the value.

**Describe the solution you'd like**
When creating a number (integer/float) feature in an annotation layer the user can set a lower and/or an upper boundary to a number feature which limits the entered values.

**Additional context**
inception-project/inception#1213

PR #1461, opened 2019-09-24
#1460 - Boundary for number features

wpmp:
**What's in the PR**
- add ranges option for number features in the settings UI

**How to test manually**
* Create a new number feature for a span annotation
* Set Limited
* Set minimum and maximum values
* For integer number features use a difference of <= 11 for minimum and maximum
* See rating panel (radio buttons) in the annotation editor panel

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**
* [x] PR updates documentation


reckart:

reckart:
I can see some issues:

* minimum can be configured to be larger than maximum - an option might be to switch the values on save if min is larger than max
* the number input field for integer values display numbers behind the comma - it does not allow entering a non-integer value though. The values should be displayed as integer values (no comma).
* when entering values on the annotation editor page, the limits do not seem to have any effect (mind that I configured the min value to be larger than the max value in the feature settings).

wpmp:
>     * minimum can be configured to be larger than maximum - an option might be to switch the values on save if min is larger than max

Yes, I thought about that today, too :)

>     * the number input field for integer values display numbers behind the comma - it does not allow entering a non-integer value though. The values should be displayed as integer values (no comma).

Does this behavior only occur in this PR for you?
Because for me this was also previously the case.
But of course I can see what I can do about that.

>     * when entering values on the annotation editor page, the limits do not seem to have any effect (mind that I configured the min value to be larger than the max value in the feature settings).

Yes, this PR is not finished, the functionality is not implemented yet.


I'm also working on a new feature editor for some kind of rating.
I will probably use selectboxes and render it like this if the number feature is an integer, a range and the range is not larger than 5 (or so).
Or would you rather have an option to choose if want your number rendered as rating?
Or even a new rating feature that is independent from the number feature?

reckart:
> Does this behavior only occur in this PR for you? Because for me this was also previously the case. But of course I can see what I can do about that.

I'm talking about the min/max fields that appear in the feature setting when you enable "limited". It may also apply to the feature editors on the annotation page - I didn't check specifically if the issue also affects these.

reckart:
> I will probably use selectboxes and render it like this if the number feature is an integer, a range and the range is not larger than 5 (or so).

I think if the range is small, using a slider in addition (or instead) of an input field may be a good idea (http://7thweb.net/wicket-jquery-ui/slider/InputSliderPage?13).

wpmp:
> I think if the range is small, using a slider in addition (or instead) of an input field may be a good idea (http://7thweb.net/wicket-jquery-ui/slider/InputSliderPage?13).

Sorry, I didn't mean selectboxes but radio buttons from kendo ui.
I'll send a screenshot when I'm back at home, I think it would be pretty nice as a score or for priority.

reckart:
There seems to be a rating panel in Wicket: org.apache.wicket.extensions.rating.RatingPanel

wpmp:
> There seems to be a rating panel in Wicket: org.apache.wicket.extensions.rating.RatingPanel

Yes, I already tried using it, however there seems to be a problem with the annotation detail editor because RatingPanel does not inherit not of type FormComponent.
I don't know if there is a way around it.

wpmp:
rating
What do you think about this?

reckart:
The idea is good :) I find it a bit difficult to see which radio button is for which rating. Haven't tried it in action, so I wonder if/how it reacts to keyboard inputs (i.e. moving focus around using tab, pressing enter to save the value).

wpmp:
I'd like to have the labels above or below the according radio button, however I still need to figure out how I could achieve that.

wpmp:
@reckart can you please have a look again?
I implemented the radio buttons.
However I needed to use events to save the changes made in the RadioGroup.
I don't know if it is possible without events but I didn't mange to get it working otherwise.

I then introduced a FeatureUpdatedEvent which can be reused for other features if needed.
For example I replaced the LinkFeatureSetEvent with the FeatureUpdatedEvent because it has the same behavior.

The radio buttons are not usable with the keyboard (arrow keys, enter) however I thought the link features aren't either.

The RadioGroup renders if the feature is of type number with integer and if the difference between minimum and maximum is <= 11 (to allow for example ratings like 0-10).

For testing with INCEpTION this requires inception-project/inception#1410 because of the changes to the events.

What do you think?

wpmp:
Ah, and I don't know how to make the NumberTextFields do not display floating input if it uses only integers.
But as I said it seems to me that this is always the case

reckart:
Controlling the number of decimals should work by passing an  `Options` object to the `NumberTextField` constructor which has `decimals` set to `0`. 

Cf. https://docs.telerik.com/kendo-ui/api/javascript/ui/numerictextbox/configuration/decimals

reckart:
We already have a `FeatureValueUpdatedEvent` which is sent by the `TypeAdapter_ImplBase` when a new feature value has been written to the CAS. So maybe the `FeatureUpdatedEvent` should be renamed to something like `FeatureEditorValueChangedEvent`?

wpmp:
Thanks, `options.set("format", "n0");` does the trick for integers. :)

wpmp:
This PR is ready for a review

reckart:

reckart:
I can still set the minimum to be smaller than the maximum in the traits.

reckart:
I think we need to make the radio-button presentation optional. I.e. the default even for small ranges should be the input field and the radio button presentation should only be shown if the project manager turns it on. One of the reasons is that the radiobutton version breaks keyboard interaction and focussing (e.g. the component returned by `RatingFeatureEditor. getFocusComponent()` doesn't seem to have a stable ID and/or is not focusable).

reckart:

wpmp:
> I can still set the minimum to be smaller than the maximum in the traits.

I think you mean you can set the maximum smaller than the minimum, right?
I fixed it.

I'll add an option to turn on the rating.

reckart:
Best add a dropdown where one can select between something like "Spinner" and "Radio buttons" - that would make it easy(ier) to later add more/other visualizations.

wpmp:
@reckart done

reckart:

wpmp:
This shouldn't be marked as WIP anymore

reckart:

reckart:

Issue #1468, opened 2019-10-03
Curatable documents are not shown


UWinch:
**Describe the bug**
The OpenDocumentDialog for the curation page does not show any documents.

**To Reproduce**
1. Have some finished docuemnts
2. Open the curation page as a curator
3. The OpenDocumetnDialog is shown
3. The project is shown but not the documents

**Expected behavior**
Should show curatable documents.

**Please complete the following information:**
 - Version and build ID: 4.0.0.-SNAPSHOT
 - OS: Ubuntu
 - Browser: Chrome

PR #1470, opened 2019-10-05
#1468 - Curatable documents are not shown

reckart:
Fixed by returning current user in user selection.

**What's in the PR**
* closes https://github.com/webanno/webanno/issues/1468

**How to test manually**
* see issue discription


reckart:
I have a hunch that correction and automation modes may also be affected... checking...

Issue #1483, opened 2019-10-18
Crash when modifying annotations while slot is armed


wpmp:
**Describe the bug**
If a slot in a link feature is armed and then some other feature in an opened annotation is modified the application crashes.

**To Reproduce**
1. Open a span annotation
2. Arm a slot in a link feature
3. Modify another feature like a string or number feature.

**Expected behavior**
The slot should get unarmed and the modification to the other feature should take place.

PR #1484, opened 2019-10-18
#1483 - Crash when modifying annotations while slot is armed

wpmp:
**What's in the PR**
- clear slot and send RenderSlotsEvent when creating or updating a span annotation and send RenderSlotsEvent
- generated new random serialVersionUID for rating feature

**How to test manually**
1. Open a span annotation
2. Arm a slot in a link feature
3. Modify another feature like a string or number feature.
4. Slot should be disarmed and the other feature should be updated

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**


reckart:
Why is it necessary to unarm the slot?

reckart:
The armed slot is part of the selection. The selection doesn't change when a feature of the selected annotation is edited - so I'd just leave it armed.

wpmp:
To be honest I would rather unarm it.
I think when you want to create a link and arm the slot you should immediately select an annotation.
If you start doing something else, like changing feature values, etc. I'd stop the process of linking annotations.
I'd always abort linking when you start doing something else than selecting an annotation.

wpmp:
The initial problem here is that when you start changing feature values while a slot is armed the list of `LinkWithRoleModel` seems to lose the new entry and we get index out of bounds in `LinkFeatureEditor` line 297.
However the editor still shows an armed slot in the UI.

reckart:

PR #1504, opened 2019-10-27
Regex Validation of string features - Extension to INCEpTION

NarraVivek:
**What's in the PR**
Added Regex Validation field in layers configuration settings page and
also added the validation based on regex in PrimitiveUimaFeatureSupport.

An attempt to solve the issue was made using the discussions present in the pull request present in the INCEpTION. Reference to that pull request is: https://github.com/inception-project/inception/pull/1488

Contribution towards the INCEpTION or WebAnno application was not completed as this task is a part of our course.


reckart:
Thanks for the PR :)

Since regex validation does not apply to *all* features, it should not be added to `AnnotationFeature` and `FeatureDetailForm` but rather to `UimaStringTraits` and `UimaStringTraitsEditor` - as discussed [here](https://github.com/inception-project/inception/issues/1268#issuecomment-544501676).

Issue #1514, opened 2019-11-01
Sessions are never cleared


reckart:
**Describe the bug**
Spring security sessions are never cleared from registry.

**Expected behavior**
When a user logs out manually or when the HTTP session expires, it should be removed from the Spring Security session registry

**Please complete the following information:**
 - Version and build ID: 3.6.2

PR #1515, opened 2019-11-01
#1514 - Sessions are never cleared

reckart:
**What's in the PR**
- Listen to sessions being created and closed and update the registry accordingly
- Count the number of unique users per PING period as active users, not just ones logged in when the PING is sent

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**

### DIFF
--- a/webanno-security/src/main/java/de/tudarmstadt/ukp/clarin/webanno/security/session/SessionListener.java
+++ b/webanno-security/src/main/java/de/tudarmstadt/ukp/clarin/webanno/security/session/SessionListener.java
@@ -1,0 +1,92 @@
+/*
+ * Copyright 2019
+ * Ubiquitous Knowledge Processing (UKP) Lab
+ * Technische Universität Darmstadt
+ *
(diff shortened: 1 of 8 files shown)


webanno-security/src/main/java/de/tudarmstadt/ukp/clarin/webanno/security/session/SessionListener.java
```diff
@@ -0,0 +1,92 @@
...
+ * login page usually takes care of registering the session.
+ */
+@Component
+public class SessionListener
+{
+    private final Logger log = LoggerFactory.getLogger(getClass());
+    
+    private final SessionRegistry sessionRegistry;
+
+    public SessionListener(SessionRegistry aSessionRegistry)
```
jcklie: Why doesnt this need `@autowired`?

reckart: Good question. It seems to work. Probably because there is only one constructor.

Issue #1583, opened 2020-01-13
Move feedbackpanel to front


UWinch:
**Is your feature request related to a problem? Please describe.**
We added a functionality in INCEpTION (https://github.com/inception-project/inception/issues/1270) that adds a dark overlay to the window. However, feedback, especially errors, should still be seen. Therefore, I would like the feedbackpanel to move to the front.

**Describe the solution you'd like**
Change css with high z-level.

PR #1584, opened 2020-01-13
#1583 Move feedbackpanel to front

UWinch:
**What's in the PR**
* Move feedbackpanel to the front via high css z-level.

**How to test manually**
* Check that feedbackpanel still looks the same.

### DIFF
--- a/webanno-ui-core/src/main/java/de/tudarmstadt/ukp/clarin/webanno/ui/core/page/ApplicationPageBase.html
+++ b/webanno-ui-core/src/main/java/de/tudarmstadt/ukp/clarin/webanno/ui/core/page/ApplicationPageBase.html
@@ -47,7 +47,7 @@
     </wicket:child>
   </div>
 
-  <div wicket:id="feedbackPanel" class="feedback-panel" style="z-index: 9999"></div>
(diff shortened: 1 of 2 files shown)


webanno-ui-core/src/main/java/de/tudarmstadt/ukp/clarin/webanno/ui/core/page/ApplicationPageBase.html
```diff
@@ -47,7 +47,7 @@
     </wicket:child>
   </div>
 
-  <div wicket:id="feedbackPanel" class="feedback-panel"></div>
+  <div wicket:id="feedbackPanel" class="feedback-panel" style="z-index: 9999"></div>
```
reckart: How about moving the z-level into the `.feedback-panel` class in `custom.less`?

UWinch: Makes sense, will do.

reckart:
@UWinch you should be able to merge this once Jenkins is good.

Issue #1594, opened 2020-01-21
Fix left sidebar when called from url


UWinch:
**Is your feature request related to a problem? Please describe.**
When calling the annotationpage from an url (e.g. from Inception recent activities link), the tabs on the left sidebar might not be valid if they are created depending on a custom `applies` function (e.g. relevant for the `curationsidebar` in Inception).

**Describe the solution you'd like**
This is because the page is constructed before the url parameters are processed. Therefore, I would like to re-initialize the list of tabs after the parameters are processed.

**Describe alternatives you've considered**

**Additional context**
For Inception https://github.com/inception-project/inception/pull/1257

PR #1596, opened 2020-01-22
#1594 Fix left sidebar when called from url

UWinch:
**What's in the PR**
* Refresh left sidebar tabs after url params are processed for the first time.

**How to test manually**
* Test that sidebar still looks good and annotation process is still smooth.

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**

### DIFF
--- a/webanno-ui-annotation/src/main/java/de/tudarmstadt/ukp/clarin/webanno/ui/annotation/AnnotationPage.java
+++ b/webanno-ui-annotation/src/main/java/de/tudarmstadt/ukp/clarin/webanno/ui/annotation/AnnotationPage.java
@@ -133,6 +133,7 @@ public class AnnotationPage
     private WebMarkupContainer actionBar;
     private AnnotationEditorBase annotationEditor;
     private AnnotationDetailEditorPanel detailEditor;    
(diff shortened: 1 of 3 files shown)


webanno-ui-annotation/src/main/java/de/tudarmstadt/ukp/clarin/webanno/ui/annotation/AnnotationPage.java
```diff
@@ -295,7 +297,7 @@ private void createAnnotationEditor(IPartialPageRequestHandler aTarget)
 
     private SidebarPanel createLeftSidebar()
     {
-        SidebarPanel leftSidebar = new SidebarPanel("leftSidebar", getModel(), detailEditor, () -> 
+        leftSidebar = new SidebarPanel("leftSidebar", getModel(), detailEditor, () ->
```
jcklie: Why do you assign `leftSidebar` here and in line 209?

UWinch: Thx, good catch, just being too hasty.

Issue #1598, opened 2020-02-03
ApplicationEvent when annopage is created


UWinch:
**Is your feature request related to a problem? Please describe.**
For one of the INCEpTION PRs, I will need to be notified when the annotationpage is created i.e. to imply that the user opened this page via dashboard or url.

**Describe the solution you'd like**
Publish an applicationevent.

**Describe alternatives you've considered**
Cannot add this as a Metadata to the annopage from INCEpTION because editorextension needs the information which has no knowledge of the page.

**Additional context**
https://github.com/inception-project/inception/pull/1257/

reckart:
It sounds strange that the backend would need to listen to whether something in the UI was created. Can you explain this in more detail

UWinch:
I have an annotationeditorextension that should only render some suggestions after the corresponding sidebar has been opened but not directly when the annotationpage is opened from either the dashboard or an url. This means we must somehow find out when the user leaves the page or how I want to do it here, when the user freshly opens the annotation page.
We thought about saving a state as metadata to the page, this would tell other UI elements if the suggestions should be shown but not the annotationeditorextension, because it has no knowledge of the annotation page and its metadata.

reckart:
@UWinch is this change still on the table?

PR #1599, opened 2020-02-03
#1598 ApplicationEvent when annopage is created

UWinch:
**What's in the PR**
* Event will be published when annotation page is created.

**How to test manually**
* Annotate a bit and switch between pages to test that loading/creating the annopage is still working.


UWinch:
Can be reviewed and merged

UWinch:
I have an annotationeditorextension that should only render some suggestions after the corresponding sidebar has been opened but not directly when the annotationpage is opened from either the dashboard or an url. This means we must somehow find out when the user leaves the page or how I want to do it here, when the user freshly opens the annotation page.
We thought about saving a state as metadata to the page, this would tell other UI elements if the suggestions should be shown but not the annotationeditorextension, because it has no knowledge of the annotation page and its metadata.

reckart:
I assume this is about rendering curation stuff on the annotation page only if the curation sidebar is open. I believe we were discussing to consider curation "active" not based on whether the sidebar is open but rather have some "start/stop curation" button in the sidebar - similar to active learning in INCEpTION. So then the annotation editor extension would need to check if that curation toggle is on or off. 

Right, we can't use the page to attach that state because the backend cannot access it.

Maybe a good way would be to add a mechanism to the AnnotatorState to get/set custom flags - similar as Wicket component metadata. The sidebar could save the curation flag in the AnnotatorState and the annotation editor extension could pick it up from there.

UWinch:
I don't think that this case is too different than the `DocumentOpenedEvent` that the `RecommendationService` listens to.

However, I can also see adding to the AnnotatorState as a possibility. I just did not want to make it bigger, because it's used a lot throughout the app. However, this solution might be more reusable, if we can make it general enough.

UWinch:
Closing this because we went with a different solution (adding metadata to annotatorstate).

PR #1618, opened 2020-02-27
#1598 annotatorstate metadata

UWinch:
**What's in the PR**
* Alternative to https://github.com/webanno/webanno/pull/1599, added metadata to annotatorstate to be able to check if user started curation instead of using an event for this.

**How to test manually**
* Test if everything in webanno still works.


reckart:

PR #1619, opened 2020-02-28
Feature/1612 upgrade to bootstrap 4

reckart:
**What's in the PR**
* Upgrade to Bootstrap 4 and SCSS

**How to test manually**
* Use every possible UI of the application and check if it still works ok.

Test the following pages:
- [x] Dashboard
- [x] Annotation view
- [x] Projects view
- [x] User view
- [x] Monitoring view
- [ ] Curation 

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**


jcklie:
Build fails. This PR is large and needs extensive checking of every UI part. I would start and write down here when I checked a certain page so we can track what we tested.

reckart:
Hm, I'll check why the build fails.

reckart:
Missed some dependencies. The GH runner build fails because a dependency is no longer compatible with Java 8. I have inquired [upstream](https://github.com/l0rdn1kk0n/wicket-webjars/issues/62) if this incompatibility is intentional. Meanwhile, I'm setting Java 11 as the minimal compatibility level for WebAnno.

jcklie:
In the annotation view, if you hover over the page counter, you can increment it (I dont think that makes sense). It would be maybe also better to have grey space inbetween the buttons or group left and right. The border thickness is different between the middle and left and right.

image

The border looks too much:

image

This looks wrong:

image

This looks wrong:

image

jcklie:
Settings view:

I dont need the border around the project name here, also the font looks a bit large.

image

Overall, the font size looks a bit too large

jcklie:
I cant click on my user name, in INCEpTION that worked.

jcklie:
user view:

Checkboxes for Account enabled and show all users have different sizes

jcklie:
Is there a way to make the footer more bootstrappy?

jcklie:
The font in progress is Times New Roman it looks and does not fit with the rest

jcklie:
In the annotation document open dialog, the font sizes and fonts look off, between the panel headers and list items

jcklie:
image

Maybe we can make these labels bolder (Name, Project type, ...)

reckart:
The issue with the stylesheets not compiling when run as a JAR should be solved now.

reckart:

reckart:
image

I don't know how you got this one. In Safari 13.0.5 and Chrome 80.0.3987.122, I don't see this thick border around the layer dropdown on the annotation page.

reckart:
These two should be fixed now. The "text" field did not render ok on Chrome.

image

image

reckart:
> In the annotation document open dialog, the font sizes and fonts look off, between the panel headers and list items

The header font was intentionally small if that is what you mean -  but I can make it larger again.

reckart:
> Checkboxes for Account enabled and show all users have different sizes

Again, this was intentional, but I change it.

reckart:
> The font in Monitoring is Times New Roman it looks and does not fit with the rest

I assume you mean the font in the SVG charts. Hm, for me it is a sans-serif font. This is not related to the BS upgrade but rather to the chart rendering. I'd say best open a separate issue for this.

reckart:
> Is there a way to make the footer more bootstrappy?

Like?

reckart:
image

> In the annotation view, if you hover over the page counter, you can increment it (I dont think that makes sense). 

Why not? You can increment/reduce or enter a value into the field and then press the eye icon to jump there. Was always like that.

> It would be maybe also better to have grey space in between the buttons or group left and right. 

For me, the paging navigator is one unit.

> The border thickness is different between the middle and left and right.

Fixed.

reckart:
> Overall, the font size looks a bit too large

As far as I understood, Bootstrap 4 uses the default font size of the browser as the base size. This seems to be larger than what we had as a default before, but I think it makes sense.

reckart:
> Maybe we can make these labels bolder (Name, Project type, ...)

Non-bold labels seem to be the default for Bootstrap 4. I also though it odd at first, but after a while I got accustomed to it and I think it makes the UI feel lighter to stick with the non-bold default.

jcklie:
image

It would be nice if there was a border around the box, white in white hurts my eyes

image

I would like a border around the page counter, it is really white else

The footer in bootstrap normally does not have a hline, but has either no divider or is a different color. Ours look really cramped. There is a `footer` component from HTML, but it looks like as if we do not use it.

UWinch:
Looks good. However, the tagsets page in the settings looks off:
tagsets

reckart:
@uwinch try clearing caches and force reloading

UWinch:
> @UWinch try clearing caches and force reloading

Still have this problem, using Chrome Version 72.0.3626.121

reckart:
@jcklie do you have the same issue on the tagset panel?

jcklie:
Tageset looks ok for me

image

jcklie:
The log says

```
2020-03-02 12:23:09 WARN [admin] org.apache.wicket.markup.html.form.AutoLabelResolver - Component: [CheckBox [Component id = disableExport]] is referenced via a wicket:for attribute but does not have its outputMarkupId property set to true
2020-03-02 12:23:09 WARN [admin] org.apache.wicket.markup.html.form.AutoLabelResolver - Component: [CheckBox [Component id = anonymousCuration]] is referenced via a wicket:for attribute but does not have its outputMarkupId property set to true
2020-03-02 12:23:10 WARN [admin] org.apache.wicket.markup.html.form.AutoLabelResolver - Component: [CheckBox [Component id = overwrite]] is referenced via a wicket:for attribute but does not have its outputMarkupId property set to true
2020-03-02 12:23:11 WARN [admin] org.apache.wicket.markup.html.form.AutoLabelResolver - Component: [CheckBox [Component id = createTag]] is referenced via a wicket:for attribute but does not have its outputMarkupId property set to true
```

UWinch:
The tagsets page looks good when I run it from the the jar, but not when I run it from eclipse.

reckart:
Any errors in the browser console?

UWinch:
> The tagsets page looks good when I run it from the the jar, but not when I run it from eclipse.

Must have been an eclipse problem, I tested it in a new workspace and now it looks fine. Sorry for the extra work.

reckart:
> It would be nice if there was a border around the box, white in white hurts my eyes

I assume you mean the select input - border added.

> I would like a border around the page counter, it is really white else

I just removed the border because before you said the border makes the input look strange in comparison to the buttons left and right...? Actually, the group adds a border around it and around all the buttons. Not sure why an extra border would be necessary - in particular since I just remove it per request.

jcklie:
I mean there never was a border between the left right and number. That blends in each other a lot. The border I meant was that the bottom and top border of the increment was different size than of the pagination.

reckart:
> I mean there never was a border between the left right and number. That blends in each other a lot. The border I meant was that the bottom and top border of the increment was different size than of the pagination.

How about you make an update to the PR illustrating the change you would like to see?

jcklie:
75554920-8fc76080-5a3b-11ea-88d0-5b7ce6af3e6f

The top and down border of the page number is different than of the pagniation, I did not fix it in the mockup

reckart:
This is with left/right border and without top/bottom border.

Screenshot 2020-03-03 at 16 53 42

jcklie:
I love it.

reckart:
Anybody still want's to double-check curation before merging this?

UWinch:
I checked it, seems good.

Issue #1694, opened 2020-05-23
Improve annotation detail sidebar architecture and performance


reckart:
**Describe the refactoring action**
The right sidebar consists of a very small number of classes and even less HTML files. These should be split up and disentangled.

**Expected benefit**
This should make the code more maintainable and also to facilitate updating only those parts of the sidebar which really require re-rendering after a particular action.

PR #1695, opened 2020-05-25
#1694 - Improve annotation detail sidebar architecture and performance

reckart:
**What's in the PR**
* Refactored the annotation detail editor on the right side
* Sidebar does not completely re-draw, but only parts that are required. This fixes the problem that you enter something into a feature, tab over to the next feature, write something, and after a moment, what you wrote disappears.
* Also reduces the amount of data exchanged between browser and server which improves the experience on slower connections
* Upgrade to recent Wicket JQuery/Kendo fixes Chrome's auto-complete interfering with the Kendo Combobox

**How to test manually**
* Annotate :)

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**


jcklie:
When I click on existing annotations, then nothing happens in the right sidebar.

rightside

jcklie:
image

reckart:
Selection in WebAnno is not a click but a double-click.

Issue #1700, opened 2020-05-28
Option to limit number of users


UWinch:
**Is your feature request related to a problem? Please describe.**
I would like the option to limit the number of currently active users to below a threshold. After this threshold is reached, users who want to log in will not be able to do so and get a message to try at a later time again.

**Describe the solution you'd like**
Optional properties value for maximum of active users is checked when logging in. If it is reached, the current user will not be able to log in and a message will be displayed on the Login Page.

**Additional context**
Tool for INCEpTION to iteratively increase the number of allowed users on a running instance.

PR #1702, opened 2020-05-29
#1700 Option to limit number of users

UWinch:
The limit of users that are able to be logged in at the same time can be set as a property. Users above this limit will not be able to log in and receive a message.

**What's in the PR**
* Check number of active users against property on login page.

**How to test manually**
* Set the `max.users` property in the `settings.properties` file to a number.
* Try to log in with more users than the defined number.

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**
* [x] PR updates documentation


webanno-ui-core/src/main/java/de/tudarmstadt/ukp/clarin/webanno/ui/core/login/LoginPage.java
```diff
@@ -75,13 +79,23 @@
...
     
     public LoginPage()
     {
         setStatelessHint(true);
         setVersioned(false);
         
         add(form = new LoginForm("loginForm"));
+        signInBtn = new Button("signInBtn", new ResourceModel("signIn"));
+        tooManyUsersLabel = new MultiLineLabel("usersLabel", new ResourceModel("tooManyUsers"));
+        tooManyUsersLabel.setVisible(false);
```
reckart, line 94: Instead of setting the visibility of a component via `setVisible`, better use a behavior, e.g. 

```
        label.add(LambdaBehavior.visibleWhen(() -> getModelObject().getPreferences().isRememberLayer()));
```

webanno-ui-core/src/main/java/de/tudarmstadt/ukp/clarin/webanno/ui/core/login/LoginPage.java
```diff
@@ -101,6 +115,29 @@ public LoginPage()
...
+        
+        if (usersProp == null || !StringUtils.isNumeric(usersProp)) {
+            return;
+        }
+        
+        // if there will be more users logged in (with current one) than max users allowed
+        if (sessionRegistry.getAllPrincipals().size() >= Integer.parseInt(usersProp)) {
+            form.setEnabled(false);
+            tooManyUsersLabel.setVisible(true);
+            signInBtn.setEnabled(false);
```
reckart, line 131: There is also a `LambdaBehavior.enabledWhen` ;)

webanno-doc/src/main/resources/META-INF/asciidoc/admin-guide.adoc
```diff
@@ -92,3 +92,5 @@ include::{include-dir}settings_custom-header-icons.adoc[leveloffset=+1]
 include::{include-dir}settings_annotation-editor.adoc[leveloffset=+1]
 
 include::{include-dir}security-preauth.adoc[leveloffset=+1]
+
+include::{include-dir}login.adoc[leveloffset=+1]
```
reckart, line 96: How about `security-login`?

Issue #1703, opened 2020-05-30
Improve brat rendering


reckart:
**Describe the refactoring action**
Profile the rendering of the brat visualization and see what can be optimized.

**Expected benefit**
Smoother user experience.

PR #1704, opened 2020-05-30
#1703 - Improve brat rendering

reckart:
**What's in the PR**
- Render brat directly in the AJAX call instead of deferring the rendering until after the AJAX call
- Added more profiling points to the brat visualizer.js
- Changed code which expands highlights for extra-wide annotations and which avoids jumpy selection to use already known measures instead of calling getBoundingClientRect() which is slow because it forces a re-layout
- Renamed BratProperties to AnnotationEditorProperties because they are really mostly general setting not only applying to brat
- Nicer tooltips for tag / field descriptions
- Fix JavaScript being wrongly loaded via a CssHeaderItem

**How to test manually**
* Annotate with brat (try LTR and RTL modes)

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**


reckart:

Issue #1712, opened 2020-06-09
Auto logout not succeeding


UWinch:
**Describe the bug**
The bug has no consequence for Webanno, has to be tested with INCEpTION. See https://github.com/inception-project/inception/issues/1673.

**To Reproduce**
1. Change session timeout to something shorter than 30min, e.g. 2min:
  * On server: in web.xml change `<session-timeout>2</session-timeout>`
  * From .jar: set `server.servlet.session.timeout=2m` in settings.properties
2. Go to AnnotationPage and click through some documents
3. Wait for the timer to run out + 1min + 10s
3. Redirect to login page does not happen

PR #1713, opened 2020-06-09
#1712 Auto logout not succeeding

UWinch:
**What's in the PR**
Clear the timer if already set that was set before the ajax call.

The bug has no consequence for Webanno, has to be tested with INCEpTION. See https://github.com/inception-project/inception/issues/1673.

**How to test manually**
Test with Inception:
1. Change session timeout to something shorter than 30min, e.g. 2min:
  * On server: in web.xml change `<session-timeout>2</session-timeout>`
  * From .jar: set `server.servlet.session.timeout=2m` in settings.properties
2. Go to AnnotationPage and click through some documents
3. Wait for the timer to run out + 1min + 10s
3. Redirect to login page should happen

Also test, that timer will be reset to 2min if an action was performed.

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**

### DIFF
--- a/webanno-ui-core/src/main/java/de/tudarmstadt/ukp/clarin/webanno/ui/core/logout/LogoutPanel.html
+++ b/webanno-ui-core/src/main/java/de/tudarmstadt/ukp/clarin/webanno/ui/core/logout/LogoutPanel.html
@@ -50,11 +50,11 @@
               
           this.start = function() {
              const TICK_RATE = 1000; // 1 second
-             if (this.timer) {
+             if (instance.timer) {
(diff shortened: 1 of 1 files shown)


reckart:
Why does the fix not have any consequences for WebAnno? Does the endless loop bug not exist here?

UWinch:
> Why does the fix not have any consequences for WebAnno? Does the endless loop bug not exist here?

Yes, I tested it for Webanno and the endless reload does not happen. I think that's because the reload comes from listening for ajax requests which might not happen in webanno. However, I could not pinpoint the exact request that causes it.

Issue #1721, opened 2020-06-25
Agreement doesn't work anymore


floschne:
**Describe the bug**
After pulling the latest master I Agreement doesn't work anymore. I can select the Feature and the Measure and also click the calculate button. After clicking it, I get an empty result table (without an error alarm in the UI).

In the debug log (https://gist.github.com/floschne/635475bb9442a149b71d418c2f8bad3e) I can read that an NPE was thrown when trying to upgrade the CASes ( `Unable to upgrade CAS
java.lang.NullPointerException: The CAS cannot be null`). 
I'm pretty sure this is due to line AgreementPage:441 in combination with line CasStorageSession:316.

**To Reproduce**
1. Open Agreement in a project with sufficient finished documents and annotations
2. Select a Feature ( I selected NE)
3. Select any Measure
4. Click Calculate...
5. See erroneous result table

**Expected behavior**
Working Agreement results

**Screenshots**

**Please complete the following information:**
 - Version and build ID: 4.0.0 (latest master build pulled on 25th June 2020)
 - Browser: latest chrome

**Additional context**
Debug Log: https://gist.github.com/floschne/635475bb9442a149b71d418c2f8bad3e

PR #1726, opened 2020-07-06
#1721 Agreement doesn't work anymore

UWinch:
Agreement can no longer be calculated (for finished, as well as with unfinished docs).

**How to test manually**
* see Issue
* uncheck `limit to finished documents`
* check that agreement is re-calculated

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**

### DIFF
--- a/webanno-api-dao/src/main/java/de/tudarmstadt/ukp/clarin/webanno/api/dao/AnnotationSchemaServiceImpl.java
+++ b/webanno-api-dao/src/main/java/de/tudarmstadt/ukp/clarin/webanno/api/dao/AnnotationSchemaServiceImpl.java
@@ -961,9 +961,9 @@ public class AnnotationSchemaServiceImpl
         // Check if the current CAS already contains the required type system
         boolean upgradePerformed = false;
         for (CAS cas : aCasIter) {
-            CasStorageSession.get().assertWritingPermitted(cas);
             
             if (cas != null && isUpgradeRequired(cas, ts)) {
+                CasStorageSession.get().assertWritingPermitted(cas);
                 upgradeCas(cas, ts);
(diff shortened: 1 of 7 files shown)


webanno-api-dao/src/main/java/de/tudarmstadt/ukp/clarin/webanno/api/dao/AnnotationSchemaServiceImpl.java
```diff
@@ -961,9 +961,9 @@ public boolean upgradeCasIfRequired(Iterable<CAS> aCasIter, Project aProject)
         // Check if the current CAS already contains the required type system
         boolean upgradePerformed = false;
         for (CAS cas : aCasIter) {
-            CasStorageSession.get().assertWritingPermitted(cas);
             
             if (cas != null && isUpgradeRequired(cas, ts)) {
+                CasStorageSession.get().assertWritingPermitted(cas);
```
reckart, line 966: I don't think we should check if writing is permitted only if we actually detect that an update is necessary. Even if no upgrade is necessary that the system is in principle allowed to perform the write operation. Was moving this assert into the condition necessary to fix the issue?

UWinch: Yes, the cas was null at this point.

reckart: I'll separate the null check from the upgrade check. Null CASes should just be skipped.

UWinch:

Issue #1736, opened 2020-07-17
Optimize ProjectServiceImpl.listProjectUsersWithPermissions


UWinch:
**Is your feature request related to a problem? Please describe.**
The method queries the db for all usernames with a certain permission and then goes through all names and asks the db first if the user exists and then gets it.

**Describe the solution you'd like**
Use a query where we join user and permission tables instead.

PR #1737, opened 2020-07-18
#1736 Optimize ProjectServiceImpl.listProjectUsersWithPermissions

UWinch:
The method queries the db for all usernames with a certain permission and then goes through all names and asks the db first if the user exists and then gets it.

**What's in the PR**
* Using one join instead of retrieving users with additional db queries

**How to test manually**
* Check that feature which use the two `listProjectUsersWithPermissions` methods still work

**Automatic testing**
* [X] PR includes unit tests


pom.xml
```diff
@@ -1156,6 +1156,12 @@
             <dependency>com.sun.xml.bind:jaxb-core</dependency>
             <dependency>com.sun.xml.bind:jaxb-impl</dependency>
             <dependency>javax.activation:javax.activation-api</dependency>
+            <!--
```
reckart, line 1159: Please exclude them in the modules where they are used for testing - not globally. In general, we want to be able to detect them accidentally being used so we can remove them in the non-essential cases.

webanno-project/src/main/java/de/tudarmstadt/ukp/clarin/webanno/project/ProjectServiceImpl.java
```diff
@@ -117,6 +117,12 @@ public ProjectServiceImpl()
         // Nothing to do
     }
     
+    public ProjectServiceImpl(EntityManager aEntityManager)
```
reckart, line 120: Please add a comment that this signature is reserved for testing. It would probably be a good idea to take the opportunity to properly use constructor injection on this class...

webanno-project/src/test/java/de/tudarmstadt/ukp/clarin/webanno/project/ProjectServiceImplTest.java
```diff
@@ -0,0 +1,126 @@
...
+        
+        assertThat(foundUsers).hasSize(2);
+        assertThat(foundUsers).contains(beate, kevin);
+    }
+    
+    @Test
+    public void listProjectUsersWithSpecificPermissions_ShouldReturnUsers() {
+        List<User> foundUsers = sut.listProjectUsersWithPermissions(testProject, ANNOTATOR);
+        
+        assertThat(foundUsers).hasSize(2);
```
reckart, line 108: how about instead of checking for size and containment to use `containsExactly`?

webanno-project/src/test/java/de/tudarmstadt/ukp/clarin/webanno/project/ProjectServiceImplTest.java
```diff
@@ -0,0 +1,126 @@
...
+                "de.tudarmstadt.ukp.inception.curation",
+                "de.tudarmstadt.ukp.clarin.webanno.model",
+                "de.tudarmstadt.ukp.clarin.webanno.security.model" 
+    })
+    public static class SpringConfig {
+        // No content
+    }
+    
+    @Test
+    public void listProjectUsersWithPermissions_ShouldReturnUsers() {
```
reckart, line 97: Please add a test that checks that a user which has a permission in a project but does not actually have an account in the system (no entry for that user in the users table) is not returned. This is currently possible because a `ProjectPermission` does not have a foreign-key constraint on the user table.

webanno-project/src/test/java/de/tudarmstadt/ukp/clarin/webanno/project/ProjectServiceImplTest.java
```diff
@@ -0,0 +1,126 @@
+/*
```
reckart, line 1: It's great to get tests :)

Issue #1749, opened 2020-08-13
Really destroy cas when deleting cas


UWinch:
**Describe the bug**
In `CasStorageServiceImpl.deleteCas(SourceDocument aDocument, String aUsername)` the object pool for the Cas is not deleted. However, we are deleting the associated file, so the cas object should also be deleted from the pool.

**Additional context**
For https://github.com/inception-project/inception/issues/1796. For re-merge, the cas state is re-set by deleting it (on disk) and then re-creating and saving a newly merged cas. However the cas is retained in the pool and taken as a basis for the re-merged one.

reckart:
Should we not keep it checked out from the pool so we keep a lock on it to prevent a concurrent thread from doing something with that cas while e we are not done re-merging yet?

reckart:
Would it make sense to overwrite the CAS in the pool with the remerged CAS? Since we must have exclusive access to the CAS during remerge that should probably be fine?

UWinch:
Currently we have exclusive access for the deletion of the cas. Then for getting the mergecas, we try `readCas` which will give us an exception because the file is no longer on disk, then we use `createCurationCas` which uses `readAnnotationCas(aRandomAnnotationDocument,UNMANAGED_ACCESS)`. Not sure if this can be dangerous.

reckart:
We use `readAnnotationCas(aRandomAnnotationDocument,UNMANAGED_ACCESS)` because we read some user's annotation file and then store it under the name of another user (i.e the CURATION user). That must be done unmanaged so the CAS storage doesn't complain that we try to save the data under another user.

PR #1750, opened 2020-08-13
#1749 Really destroy cas when deleting cas

UWinch:
**What's in the PR**
* invalidate casholder object in exclusiveaccesspool when deleting a cas
* call method that removes cas from pools when re-merging

**How to test manually**
* Test that re-merging during curation will reset the curation document again
* Check annotation process for buggy behaviour

### DIFF
--- a/webanno-api-dao/src/main/java/de/tudarmstadt/ukp/clarin/webanno/api/dao/CasStorageServiceImpl.java
+++ b/webanno-api-dao/src/main/java/de/tudarmstadt/ukp/clarin/webanno/api/dao/CasStorageServiceImpl.java
@@ -740,7 +740,7 @@ public class CasStorageServiceImpl
             sharedAccessCache.invalidate(key);
             
             try {
-                exclusiveAccessPool.clear(new CasKey(aDocument, aUsername));
+                exclusiveAccessPool.clear(key);
                 access.invalidateOnClose();
             }
             catch (Exception e) {
--- a/webanno-api-dao/src/main/java/de/tudarmstadt/ukp/clarin/webanno/api/dao/CasStorageServiceImpl.java
+++ b/webanno-api-dao/src/main/java/de/tudarmstadt/ukp/clarin/webanno/api/dao/CasStorageServiceImpl.java
@@ -736,10 +736,12 @@ public class CasStorageServiceImpl
             boolean success = new File(getAnnotationFolder(aDocument), aUsername + ".ser").delete();
             
             // Drop the CAS from the shared CAS it doesn't ghost around
-            sharedAccessCache.invalidate(new CasKey(aDocument, aUsername));
+            CasKey key = new CasKey(aDocument, aUsername);
+            sharedAccessCache.invalidate(key);
             
             try {
                 exclusiveAccessPool.clear(new CasKey(aDocument, aUsername));
+                access.invalidateOnClose();
             }
             catch (Exception e) {
                 throw new IOException(e);
@@ -913,6 +915,7 @@ public class CasStorageServiceImpl
         private String documentName;
         private long documentId;
         private String username;
+        private boolean invalidateOnClose;
         
         public WithExclusiveAccess(SourceDocument aDocument, String aUser)
             throws CasSessionException
@@ -936,7 +939,12 @@ public class CasStorageServiceImpl
                 holder = null;
             }
         }
-        
+
+        public void invalidateOnClose()
+        {
+            invalidateOnClose = true;
+        }
+
         public CasKey getKey()
         {
             return key;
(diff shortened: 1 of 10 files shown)


webanno-api-dao/src/main/java/de/tudarmstadt/ukp/clarin/webanno/api/dao/CasStorageServiceImpl.java
```diff
@@ -999,9 +1007,17 @@ public void close()
...
             if (holder != null) {
                 try {
-                    log.trace("Returning briefly borrowed CAS [{}]@[{}]({})", username,
-                            documentName, documentId);
-                    exclusiveAccessPool.returnObject(key, holder);
+                    if (invalidateOnClose) {
+                        log.trace("Invalidating CAS [{}]@[{}]({})", username,
+                                documentName, documentId);
+                        exclusiveAccessPool.invalidateObject(key, holder);
+                        exclusiveAccessHolders.remove(holder);
```
reckart, line 1014: Why do we need to remove the holder? According to the comment in `CasHolder borrowCas(CasKey aKey)`, that should not be necessary:

```
            CasHolder holder = exclusiveAccessPool.borrowObject(aKey);
            // Add the holder to the set of known holder. Because this set it using weak
            // references, and because we use the set only to inform holders when they become
            // invalid we do never have to explicitly remove the holder from the set
            exclusiveAccessHolders.add(holder);
```

webanno-api-dao/src/main/java/de/tudarmstadt/ukp/clarin/webanno/api/dao/CasStorageServiceImpl.java
```diff
@@ -999,9 +1007,17 @@ public void close()
         {
             if (holder != null) {
                 try {
-                    log.trace("Returning briefly borrowed CAS [{}]@[{}]({})", username,
-                            documentName, documentId);
-                    exclusiveAccessPool.returnObject(key, holder);
+                    if (invalidateOnClose) {
+                        log.trace("Invalidating CAS [{}]@[{}]({})", username,
+                                documentName, documentId);
+                        exclusiveAccessPool.invalidateObject(key, holder);
```
reckart, line 1013: Why do we need to call `exclusiveAccessPool.invalidateObject(key, holder);` here if it is already called here?

https://github.com/webanno/webanno/pull/1750/files#diff-71ecaac6a6015990de0c535cc226f8b4R743-R744

webanno-curation/src/main/java/de/tudarmstadt/ukp/clarin/webanno/curation/storage/CurationDocumentService.java
```diff
@@ -67,19 +67,6 @@ CAS readCurationCas(SourceDocument document)
...
-     * Remove a curation annotation document from the file system, for this {@link SourceDocument}
-     *
-     * @param sourceDocument
-     *            the source document.
-     * @param username
-     *            the username.
-     * @throws IOException
-     *             if an I/O error occurs.
-     */
-    void removeCurationDocumentContent(SourceDocument sourceDocument, String username)
```
reckart, line 81: From my understanding, `removeCurationDocumentContent` was used here exactly because it does not remove the CAS from the session but only clears the underlying file. The idea was (probably) that we keep a lock on the CAS in the session to avoid a concurrent thread from obtaining a lock and doing something with the CAS (e.g. regenerating it) while we are still in the re-merge process.

webanno-api-dao/src/main/java/de/tudarmstadt/ukp/clarin/webanno/api/dao/CasStorageServiceImpl.java
UWinch: I'm setting a flag to call invalidate when closing the exclusiveaccess. How I understand it: with `clear` in the delete method we will only delete idle objects and the cas we are trying to delete is still allocated because we borrowed it with the exclusiveaccess. But if we invalidate it at the same time in the delete method, we will have problems when trying to put back on access.close. So, it made the most sense to me to tell the access with the flag to invalidate the object on close.

webanno-api-dao/src/main/java/de/tudarmstadt/ukp/clarin/webanno/api/dao/CasStorageServiceImpl.java
UWinch: You are right, it should not be necessary. it made the most sense to me at the time because this way we really delete all cas instances of cases that no longer exist. Otherwise we still have an outdated version in the holder.

webanno-api-dao/src/test/java/de/tudarmstadt/ukp/clarin/webanno/api/dao/CasStorageServiceImplTest.java
```diff
@@ -93,6 +93,9 @@ public void testWriteReadExistsDeleteCas() throws Exception
         sut.deleteCas(doc, user);
         assertThat(sut.getCasFile(doc, user)).doesNotExist();
         assertThat(sut.existsCas(doc, user)).isFalse();
+        
+        // check that cas is no longer in the active session
+        assertThat(casStorageSession.contains(cas.getCas())).isFalse();
```
UWinch, line 98: @reckart this test is currently failing because we have neither set a casholder in the session or in the withexclusiveacess. do you know if cases like this exist in real life?

reckart: This test was asserting against the wrong CAS - against the original template CAS, not against the CAS that was read (into the session) and then deleted.

UWinch (commented):
Looks good, thanks. Tested deleting the cas via remote api calls to delete annotation doc and curation doc.

reckart:
Would you like to approve? ;) @UWinch

UWinch:
Github won't let me approve my own PR. However, i fully approve :)

reckart:
Yeah.... kind of a "bug" in the review concept that if the reviewer joins the development, the original submitter can not approve of the reviewer.

reckart (approved):

Issue #1755, opened 2020-08-14
Support definition of constraints across slot features


reckart:
**Is your feature request related to a problem? Please describe.**
When using slot features, I might care to define a constraint where the condition needs to look at the slot fillers. E.g. **If there is one named entity with value "PER" filling and another named entity with value "ORG" in the slots, then I want to restrict the category of the host span to "buy". 

**Describe the solution you'd like**
E.g. something like this should work:

```
import webanno.custom.EventTrigger as TRIGGER;

TRIGGER {
  participants.target.value = "PER" & participants.target.value = "LOC" -> value = "move";
  participants.target.value = "PER" & participants.target.value = "ORG" -> value = "buy";
  participants.target.value = "ORG" & participants.target.value = "ORG" -> value = "buy";
  participants.target.value = "PER" & participants.target.value = "PER" -> value = "buy";
}
```

**Describe alternatives you've considered**
It could be useful to also allow conditions on path steps such as:

```
participants[role="buyer"].target.value = "ORG"
```

**Additional context**
* See also: https://github.com/webanno/webanno/issues/1746

PR #1756, opened 2020-08-14
#1755 - Support definition of constraints across slot features

reckart:
**What's in the PR**
- Added support for constraint conditions to work across multi-valued features with other FSes as targets (i.e. slot features)
- Cleaned up code
- Upgraded JavaCC Maven plugin
- Added AnnotationBuilder and FeatureStructureBuilder helper classes in support module (quite useful for testing at least)

**How to test manually**
* Define a new span layer `EventTrigger` with
  * a slot feature called `participants` that targets the **Named entity** layer
  * a string feature called `value` which has a tagset with the tags `move` and `buy`
* Upload a document with e.g. text `Oracle bought Sun.`
* Annotate **Oracle** and **Sun** as **Named entities**, both with the value **ORG**
* Annotate **bought** as an **EventTrigger** and fill the named entities into its participant slots (role does not matter)
* In the project settings, import the following constraints:
```
import webanno.custom.EventTrigger as TRIGGER;

TRIGGER {
  participants.target.value = "PER" & participants.target.value = "LOC" -> value = "move";
  participants.target.value = "PER" & participants.target.value = "ORG" -> value = "buy";
  participants.target.value = "ORG" & participants.target.value = "ORG" -> value = "buy";
  participants.target.value = "PER" & participants.target.value = "PER" -> value = "buy";
}
```
* Go back to the annotation page, select the **EventTrigger** annotation and check that the **value** feature dropdown lists **buy** in bold at the top.

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**


UWinch (commented):
Works great with previous examples for constraints and work with the example for *Buy*. However when I annotate an example PER -> move -> LOC, I'm getting both suggestions *Move* and *Buy* i.e. both are in bold font in the dropdown. Can you check again?
Btw I used this opportunity to add a simple example for constraints in the user guide.

reckart:
@UWinch - that is not a bug in the code but in the example rules given above. The *buy* is added by the following (pointless) rule:

```
participants.target.value = "PER" & participants.target.value = "PER" -> value = "buy";
```

reckart:
Thanks for the docs.

UWinch:
> @UWinch - that is not a bug in the code but in the example rules given above. The _buy_ is added by the following (pointless) rule:
> 
> ```
> participants.target.value = "PER" & participants.target.value = "PER" -> value = "buy";
> ```

Why would this rule match a participant of "PER" and one of "LOC" though?
Screenshot from 2020-08-30 10-34-37

Issue #1757, opened 2020-08-14
Re-Merge does not work anymore


reckart:
**Describe the bug**
See https://github.com/inception-project/inception/issues/1796 for details

PR #1758, opened 2020-08-14
#1757 - Re-Merge does not work anymore

reckart:
**What's in the PR**
- Add option to force-recreate the merge CAS when loading it
- Remove upgrade for all finished annotation documents when loading the merge CAS - IMHO that should not be necessary anymore
- Instead of deleting the curation CAS on disk and then re-create it implicitly (which currently does not work because we simply get the copy from memory when we try this), just create a new merge CAS and then overwrite the curation CAS with the new merged version.

**How to test manually**
* See (upstream) issue description

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**


reckart:
@UWinch alternative take on https://github.com/webanno/webanno/issues/1749 - WDYT?

Issue #1759, opened 2020-08-16
User documents view disappear on curation page


UWinch:
**Describe the bug**
On the curation page when part of the page is re-loaded, the views showing the users' annotations are not shown.

**To Reproduce**
1. Open a document on the curation page
2. Cause a partial re-load e.g. by changing the number of displayed rows or re-merging the annotations
3. See that the users' documents are no longer shown, can be retrieved by reloading the page

**Expected behavior**
Should not disappear.

**Screenshots**
![curation-missing-views](https://user-images.githubusercontent.com/29500351/89536121-b4e4aa00-d7f7-11ea-8c30-a7bb34c7fc97.png)

**Please complete the following information:**
 - Version and build ID: 0.17.0-SNAPSHOT
 - OS: Ubuntu
 - Browser: Chrome

**Additional context**
Found in Inception https://github.com/inception-project/inception/issues/1794

PR #1760, opened 2020-08-16
#1759 User documents view disappear on curation page

UWinch:
User segments were not updated when reloading the SuggestionViewPanel after re-merge or changing preferences. Most likely, because there was an update request when the user segments no longer existed. After  researching the update process of the CurationPage and its components, I've come to the conclusion that it would need some extensive re-factoring (of components, their models and when they are updated) to fix this via the update processes.
So, instead we are using try-catch around the js calls.

**What's in the PR**
* Delay update request
* Added try-catch block to js 

**How to test manually**
* see issue
* Test overall functionality to make sure the app is still working as desired.

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**


reckart:
Ok, this probably fixes the error that shows up in the Wicket AJAX console - but does it fix the problem that the annotators views disappear?

UWinch:
> Ok, this probably fixes the error that shows up in the Wicket AJAX console - but does it fix the problem that the annotators views disappear?

Oh, it fixed it for me. Did it not work for you when testing?

reckart:
Admittedly, I didn't test it yet. But from looking at the change, it only seems to disable rendering if the element ID is not found - I didn't quite understand why disabling the rendering would fix the rendering. Can you explain?

UWinch:
To my understanding, at the point of the error, we are getting one rendering request for the views with ids that do not exist any longer (My guess when clearing the selection upon opening the preferences or re-merge dialog). This is the one we are no longer rendering with this fix. After this, we are updating the views with the new ids upon closing the the dialog again. This one always existed and does render. It must have been blocked by the error before, my guess is that's why using `setTimeout(renderCmd, 0)` also rendered the views but did not remove the error message (there was still a null pointer).

reckart:
That means that doing the rendering in a JavaScript try-catch block should also fix the issue? I ask because if that is the case, we should probably check all places where we use JavaScript to ensure that we put all calls into try-catch-blocks to avoid one such block failing the subsequent ones.

UWinch:
The error already occurs because we cannot find the element at `Wicket.$('" + vis.getMarkupId() + "').dispatcher.post('renderData', [" + aJson+ "]);"` because the element is null and we are calling .dispatcher on null. The renderData js method already uses try-catch. But, we could put calls to js into try-catches as well. For this case e.g.:

```
private String bratRenderCommand(String aJson)
    {
        return "try{" +
                "  Wicket.$('\" + vis.getMarkupId() + \"').dispatcher.post('renderData', [" + 
                   aJson + "]);" +
                "}catch (e){" +
                "  console.warn('Rendering call terminated due to: ' + e, e.stack);" +
                "}";
    }
```

reckart:
Right. And if this helps, we should probably scan the whole codebase for places where we attach JavaScript snippets to an AJAX target and add try/catch in all places - maybe via some helper method we could introduce in the support module which simply wraps the given command in a try/catch and logs the error if an error occurs.

webanno-ui-core/src/main/java/de/tudarmstadt/ukp/clarin/webanno/ui/core/page/webanno.js
```diff
@@ -39,3 +39,11 @@ $(document)
       }
     });
 
+//wrap given script in try-catch block
+function tryCatch(jsCall){
+	try{
+		jsCall();
+	}catch (e){
+		console.warn('Call terminated due to: ' + e + ' script was:\n' + jsCall + '\n', e.stack);
```
reckart, line 47: I think trying to print "jsCall" won't print the script but rather some representation of the function. Did you test this?

webanno-ui-core/src/main/java/de/tudarmstadt/ukp/clarin/webanno/ui/core/page/webanno.js
```diff
@@ -39,3 +39,11 @@ $(document)
       }
     });
 
+//wrap given script in try-catch block
+function tryCatch(jsCall){
+	try{
```
reckart, line 44: Formatting (spaces around braces)

webanno-ui-curation/src/main/java/de/tudarmstadt/ukp/clarin/webanno/ui/curation/component/SuggestionViewPanel.java
```diff
@@ -135,7 +134,7 @@
 
     private static final long serialVersionUID = 8736268179612831795L;
 
-    private static final Logger LOG = LoggerFactory.getLogger(SuggestionViewPanel.class);
+    private static final Logger log = LoggerFactory.getLogger(SuggestionViewPanel.class);
```
reckart, line 137: static final stuff should be in CAPITALS

webanno-ui-curation/src/main/java/de/tudarmstadt/ukp/clarin/webanno/ui/curation/page/CurationPage.java
```diff
@@ -827,13 +826,10 @@ public void onClick(AjaxRequestTarget aTarget)
                 CAS cas = curationDocumentService.readCurationCas(state.getDocument());
                 updateCurationView(curationContainer, curationViewItem, aTarget,
                         cas);
-                updatePanel(aTarget, curationContainer);
+//                updatePanel(aTarget, curationContainer);
```
reckart, line 829: Why is it not needed anymore? Why not entirely remove it instead of commenting out?

UWinch: `updateCurationView` calls `updatePanel`. I only commented it out though in case I missed something and we find that this messes something up upon more testing

reckart (approved):

PR #1766, opened 2020-09-01
Feature/1765 clean up constraints grammar file

reckart:
**What's in the PR**
- Switch from JTB to JJTree

**How to test manually**
* Try out some of the constraints examples from the user manual

**Automatic testing**
* [x] PR includes unit tests (well, slightly updated tests at least, although some test went away due to the change from JTB to JJTree, they are no longer needed)

**Documentation**


UWinch (commented):
I cannot get the constraint examples for slot features to work e.g. https://github.com/webanno/webanno/pull/1756.
Also, I noticed that the mousover that indicates if a rule matches (on the green checkmark, yellow i) does not update its content in time i.e. the rule matches but it might still say that none of the rules matched.

reckart:
@UWinch I followed the instructions in #1756 and it worked for me. What didn't work for you?

UWinch (approved):
I just tried it again and it worked for me, not sure what the problem was before. Probably user error ;)

Issue #1764, opened 2020-08-25
Error when annotating new layer for the first time


UWinch:
**Describe the bug**
After annotating a document, then adding a new layer and annotating this annotation type for the first time, an error is thrown.
Might be that cas is not upgraded soon enough?

<details>

```
java.lang.IllegalArgumentException: Undeclared type [webanno.custom.Testing2] Available types: uima.cas.TOP...[without Testing2]...
at org.apache.uima.fit.util.CasUtil.getType(CasUtil.java:142) ~[uimafit-core-3.1.0.jar:3.1.0]
	at de.tudarmstadt.ukp.clarin.webanno.api.annotation.adapter.SpanOverlapBehavior.onCreate(SpanOverlapBehavior.java:74) ~[classes/:?]
	at de.tudarmstadt.ukp.clarin.webanno.api.annotation.adapter.SpanAdapter.handle(SpanAdapter.java:108) ~[classes/:?]
	at de.tudarmstadt.ukp.clarin.webanno.api.annotation.adapter.SpanAdapter.add(SpanAdapter.java:100) ~[classes/:?]
	at de.tudarmstadt.ukp.clarin.webanno.ui.annotation.detail.AnnotationDetailEditorPanel.createNewSpanAnnotation(AnnotationDetailEditorPanel.java:435) ~[classes/:?]
	at de.tudarmstadt.ukp.clarin.webanno.ui.annotation.detail.AnnotationDetailEditorPanel.createNewAnnotation(AnnotationDetailEditorPanel.java:398) ~[classes/:?]
	at de.tudarmstadt.ukp.clarin.webanno.ui.annotation.detail.AnnotationDetailEditorPanel.internalCommitAnnotation(AnnotationDetailEditorPanel.java:853) ~[classes/:?]
	at de.tudarmstadt.ukp.clarin.webanno.ui.annotation.detail.AnnotationDetailEditorPanel.actionCreateOrUpdate(AnnotationDetailEditorPanel.java:632) ~[classes/:?]
	at de.tudarmstadt.ukp.clarin.webanno.brat.annotation.BratAnnotationEditor.actionSpan(BratAnnotationEditor.java:471) ~[classes/:?]
	at de.tudarmstadt.ukp.clarin.webanno.brat.annotation.BratAnnotationEditor.access$5(BratAnnotationEditor.java:445) ~[classes/:?]
[...]

PR #1767, opened 2020-09-05
#1764 - Error when annotating new layer for the first time

reckart:
**What's in the PR**
- Replace the Spring ConcurrentReferenceHashMap with a simpler synchronized WeakHashMap turned into a set because the Spring class seems to "loose" entries without any good reason
- Added a test case for the type-system-changed event
- Added some trace-level logging code for debugging

**How to test manually**
* See issue description

**Automatic testing**
* [x] PR includes unit tests

**Documentation**

### DIFF
--- a/webanno-api-annotation/src/main/java/de/tudarmstadt/ukp/clarin/webanno/api/annotation/util/ThreadLockingInvocationHandler.java
+++ b/webanno-api-annotation/src/main/java/de/tudarmstadt/ukp/clarin/webanno/api/annotation/util/ThreadLockingInvocationHandler.java
@@ -17,15 +17,19 @@
  */
 package de.tudarmstadt.ukp.clarin.webanno.api.annotation.util;
 
+import static java.util.Arrays.asList;
 import static java.util.stream.Collectors.joining;
 
 import java.lang.reflect.InvocationHandler;
 import java.lang.reflect.Method;
+import java.util.List;
 import java.util.stream.Stream;
 
 public class ThreadLockingInvocationHandler
     implements InvocationHandler
 {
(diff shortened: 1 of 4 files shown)


reckart:
@UWinch looks like the problem was that `ConcurrentReferenceHashMap` uses weak references for the *keys and values* while the code (or rather me writing the code) expected only the *keys* to be weak references as in a normal `WeakHashMap`. So when we applied `newSetFromMap()` to the `ConcurrentReferenceHashMap`, we internally created a `Map<CasHolder, Boolean>` in which the `CasHolder` keys would disappear when the `Boolean` dummy value was garbage collected (weak value references). I switch the a `synchronized(newSetFromMap(WeakHashMap))` where only the keys are weak, so the `CasHolder` should now only disappear from the set when it has been garbage collected - which was the original expectation and intended behaviour when the code was conceptualised.

UWinch:
I'm still getting the error when:

- open a document first
- add a new layer via settings
- try to annotate with the new layer on the same document

Does it work for you?

reckart:
When I tested, I am pretty sure it worked for me. Maybe try cleaning/rebuilding the workspace?

reckart:
@UWinch you were right, it was still broken in INCEpTION. The issue appears to have been that immediately after adding the layer, the CASes were re-loaded into memory by the indexing thread using NO_UPGRADE. And afterwards, when we opened a document in the annotation editor, the FORCE_UPGRADE flag was ignored because the CAS was already in memory. I have changed the code to upgrade/repair CASes even if they were borrowed from the pool. That seems to fix the loophole.

UWinch (approved):
Thanks a lot for the fix.

Issue #1771, opened 2020-09-10
Curation page does not load


UWinch:
**Describe the bug**
After opening the curation page and selecting a document to curate, the loading process does not finish.

**To Reproduce**
1. Have a document finished
2. Try opening it on the curation page.

**Additional context**
Found with INCEpTION

PR #1772, opened 2020-09-10
#1771 - Curation page does not load

reckart:
**What's in the PR**
- Fixed exception being thrown as part of the logging when handling an exception while trying to load a CAS causing the borrowed CAS holder for this CAS never to be returned which in turn caused a deadlock *puh*
- Added maximum wait time for borrowing a CAS

**How to test manually**
* See issue description

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**

### DIFF
--- a/webanno-api-dao/src/main/java/de/tudarmstadt/ukp/clarin/webanno/api/dao/CasStorageServiceImpl.java
+++ b/webanno-api-dao/src/main/java/de/tudarmstadt/ukp/clarin/webanno/api/dao/CasStorageServiceImpl.java
@@ -99,6 +99,7 @@ public class CasStorageServiceImpl
 
     private final long EVICT_IDLE_CASES_AFTER_MINUTES = 5;
     private final long SHARED_CAS_CACHE_SIZE = 10_000;
+    private final long CAS_BORROW_WAIT_TIMEOUT_MINUTES = 3;
 
     private final CasDoctor casDoctor;
     private final AnnotationSchemaService schemaService;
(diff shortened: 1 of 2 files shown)


reckart:
@UWinch one of the reasons behind this issue is that the mechanism which initializes a curation CAS when there is not yet is not very well written. In `de.tudarmstadt.ukp.clarin.webanno.ui.curation.component.model.SuggestionBuilder.getMergeCas(AnnotatorState, SourceDocument, Map<String, CAS>, AnnotationDocument, boolean, boolean, boolean)`, we try to load the merge CAS (what's displayed in the upper part of the screen of the curation page) and by **ANY EXCEPTION** is thrown while we do that, we simply re-initialize the merge CAS (perform a remerge):

```
        // Create JCas, if it could not be loaded from the file system
        catch (Exception e) {
            return initializeMergeCas(aState, aCasses, aTemplate, aMergeIncompleteAnnotations);
        }
```

The idea here is that we should perform a re-merge if there  is **no merge CAS** so far. However, the present implementation simply catching Exception will also override any potentially existing merge CAS if there is any type of problem loading the CAS. Instead of doing that, it would seem to me to be more reasonable to use `correctionDocumentService.existsCorrectionCas(sourceDocument)` and `curationDocumentService.existsCurationCas(...)` (the latter would need to be implemented because it is not there yet) to check if a merge CAS exist and only if it does not exist to perform a re-merge - but if it exists and there is trouble loading it, then we should rather throw an exception saying what the trouble is instead of simply overwriting the data.

UWinch (approved):
Works again, nice find of the cause!

PR #1780, opened 2020-09-14
Feature/1775 curation cas creation

UWinch:
Merge cases for the curation and correction modes should not be (re-)initialized on error, instead when no previous cas was saved.

**What's in the PR**
* check for existing cas when getting merge cas
* ~refactor: moved common functionalities of `CorrectionDocumentService` and `CurationDocumentService` to `MergeDocumentService`~

**How to test manually**
* Check that curation still works
* Check that correction/automation still works


reckart:
Reverted factoring out common code between CurationDocumentService and CorrectionDocumentService into a parent class. It only does away with a single condition on the project mode and the inheritance introduces a kind of relationship between the CurationDocumentService and the CorrectionDocumentService that we don't want to have. Let's try this again but in a way where we focus specifically on getting rid of conditions on the project mode.

UWinch:
Ok, so do we merge this as-is now?

reckart (approved):

reckart:
I was waiting for Jenkins and in fact I haven't tried it out locally yet.

webanno-ui-curation/src/main/java/de/tudarmstadt/ukp/clarin/webanno/ui/curation/component/model/SuggestionBuilder.java
```diff
@@ -334,8 +343,9 @@ public CAS getMergeCas(AnnotatorState aState, SourceDocument aDocument,
         }
         // Create JCas, if it could not be loaded from the file system
         catch (Exception e) {
-            return initializeMergeCas(aState, aCasses, aTemplate, aMergeIncompleteAnnotations);
+            log.error(e.getMessage());
```
reckart, line 346: I think we should probably not catch exception here but rather let the exception bubble up if there is any.

reckart (approved):

Issue #1785, opened 2020-09-25
Feature Editor wrongly shown


UWinch:
**Describe the bug**
Sometimes the feature editor of previously selected annotations is still shown even though the layer is no longer selected.

**To Reproduce**
1. Select an annotation
2. Start an AL session

**Expected behavior**
When no layer is selected, then the feature editor should not be shown.

**Screenshots**
![layer_feature_error](https://user-images.githubusercontent.com/29500351/94273643-f581b980-ff44-11ea-9927-aa4a4afe9946.png)

**Please complete the following information:**
 - Version and build ID: INCEpTION 0.17.0-SNAPSHOT
 - OS: ubuntu
 - Browser: chrome

**Additional context**
Found in INCEpTION

PR #1786, opened 2020-09-25
#1785 Feature Editor wrongly shown

UWinch:
Sometimes features of previously selected annotations were still shown.

**What's in the PR**
*  Feature editor is not visible when no selected layer

**How to test manually**
* Select annotation
* Open AL session or re-merge from curations sidebar
* Right sidebar should show no layer selected and also no features


webanno-ui-annotation/src/main/java/de/tudarmstadt/ukp/clarin/webanno/ui/annotation/detail/FeatureEditorListPanel.java
```diff
@@ -92,16 +92,21 @@ public FeatureEditorListPanel(String aId, IModel<AnnotatorState> aModel,
         featureEditorContainer.setOutputMarkupPlaceholderTag(true);
         featureEditorContainer.add(featureEditorPanelContent = new FeatureEditorPanelContent("featureEditors"));
         featureEditorContainer.add(createFocusResetHelper());
-        featureEditorContainer.add(visibleWhen(() -> !hasFeatures()));
+        featureEditorContainer.add(visibleWhen(() -> selectedLayerHasFeatures()));
```
reckart, line 95: Better rename it to `layerIsSelectedAndHasFeatures()` and `layerIsSelectedButHasNoFeatures()` - that makes it clearer that neither widget is shown when no layer is selected.

reckart:
Thanks! :)

reckart (approved):

Issue #1822, opened 2020-12-05
Annotation disabled for curator


UWinch:
**Describe the bug**
There are two cases in which the user in the annotator state is not the current user. The annotation should only be disabled in one of them:

- The admin (current user) is viewing another annotator's work -> the annotation should be disabled
- The current user is acting as the curation user -> the annotation should not be disabled

**To Reproduce**
Relevant for https://github.com/inception-project/inception/pull/1872.

**Describe the solution you'd like**
Revert to the previous functionality of  `isUserViewingOthersWork`, move checks for curation to `editable`-methods.

reckart:
The check was before using the user object instead of the user name to check access. This required an additional DB access to fetch the user object which slowed down things. So we should not revert but fix without reintroducing the DB access.

PR #1823, opened 2020-12-08
#1822 Annotation disabled for curator

UWinch:
For https://github.com/inception-project/inception/pull/1872 regarding document state for curation when curating from sidebar.

**What's in the PR**
* Fixed that annotation was disabled by adding curation check to `editable`-methods
* Now tracking and transitioning curation document status in annotation mode to support curation sidebar and keep document state up to date (cf. Monitoring Page)

**How to test manually**
* Annotate, curate and finish documents and check that document state is up to date.


reckart (approved):
Did some testing in the UI. Looks ok. Thanks!

Issue #1860, opened 2021-01-19
Redirect to previous url after login does not work


UWinch:
**Describe the bug**
When I'm not logged in and want to access a specific page, I get re-directed to the login page. After logging in, I would like to be re-directed to the page I originally requested. Found in Inception with an url for a specific project dashboard.

**Please complete the following information:**
 - Version and build ID: INCEpTION 0.18.0-SNAPSHOT
 - Browser: chrome

PR #1861, opened 2021-01-19
#1860 Redirect to previous url after login does not work

UWinch:
**What's in the PR**
* Get previous url before sign in

**How to test manually**
* login via `/login` -> be directed to welcome.html page
* log out
* request e.g. a specific document `/annotation.html?3#!p=3&d=9` -> should be re-directed to this document


reckart (approved):
Works!

Issue #1869, opened 2021-02-15
Error during curation if CASes not fully upgraded


reckart:
**Describe the bug**
In some cases, there can be an exception if a curation is attempted when the CASes are not fully upgraded. This happens in particular if a feature is present in one CAS but not in another.

**To Reproduce**
Discovered during development - no specific steps to reproduce.

**Expected behavior**
There should not be an error. If a feature does not exist during comparison, a default value should be assumed based on the feature type.

**Please complete the following information:**
 - Version and build ID: WebAnno 4.0.0-beta-22

PR #1870, opened 2021-02-15
#1869 - Error during curation if CASes not fully upgraded

reckart:
**What's in the PR**
- Make sure the feature is looked up from the right type system
- Use a default value depending on the feature type if the feature itself could not be found
- Added missing returns in getFeatureValue

**How to test manually**
* No specific test procedure

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**


reckart:

PR #1914, opened 2021-10-03
Cognitive Complexity of functions should not be too high

viveshini:
What is the issue/code smell?
The cognitive Complexity of functions should not be too high.
Why is this issue relevant?
Cognitive Complexity is a measure of how hard the control flow of a function is to understand. Functions with high Cognitive Complexity will be difficult to maintain. If cognitive complexity is too high, we can say that the code is complex and very hard to maintain.
How can this issue be resolved?
The cognitive complexity of functions can be reduced by reducing the conditions and method extraction.


reckart:
This PR removes two empty lines. I don't quite see the point.

Issue #70, opened 2015-08-05
Reset/recreate default admin account


reckart:
In particular for the standalone version where the database contents are not as easily accessible, having some way to reset the password of the built-in admin account would be useful.

3.2.x: The user should be able to specify an environment variable like `webanno.restoreDefaultAdminAccount`. If that variable is present, going to the login page will reset/recreate the default admin account. At the same time, it will disable login.

3.3.x+: The user should be able to specify an environment variable like `webanno.restoreDefaultAdminAccount`. If the variable is present, the account should be reset/recreated during the application startup and then the application should be shut down again.

reckart:
Has been implemented in INCEpTION -> https://github.com/inception-project/inception/issues/2503

PR #1915, opened 2021-10-20
Reset/Recreate default admin account

lokeshmekala:
Reset/ Recreate the default admin account.
The current version of webanno software creates a default admin account when there are no users account found. When the software runs for the first time, there will not be any user accounts, Hence to allow the user to access the software, webanno creates a default admin account with the username as admin and password as admin. We have to provide the liberty to the user to create his default admin account. 
To allow the user to create his default admin account, we provide two environment variables. One of the environment variables contains the default username and the other default password(WEBANNO_ENV_ADMIN_NAME, WEBANNO_ENV_ADMIN_PASS). If these variables are present, then these values are considered to create the default admin account. If the values are not present, the default admin account is created with username as "admin" and password as "admin".

### DIFF
--- a/webanno-ui-core/src/main/java/de/tudarmstadt/ukp/clarin/webanno/ui/core/login/LoginPage.java
+++ b/webanno-ui-core/src/main/java/de/tudarmstadt/ukp/clarin/webanno/ui/core/login/LoginPage.java
@@ -101,14 +101,29 @@ public class LoginPage
         // Create admin user if there is no user yet
         if (userRepository.list().isEmpty()) {
             User admin = new User();
-            admin.setUsername(ADMIN_DEFAULT_USERNAME);
-            admin.setPassword(ADMIN_DEFAULT_PASSWORD);
+            //Checking for any env variable with the "WEBANNO_ENV_ADMIN_NAME" name, this variable is used for setting username.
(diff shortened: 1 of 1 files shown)


reckart:

reckart:
The title of the PR is "Reset/Recreate default admin account". However, setting the variables does not in any way reset or recreate the admin account. They only change the username and password of the default admin account that is created when no user exists in the database. 

So it looks to me like the PR does not meet the goal of resetting/recreating the default admin account.

There is an issue about resetting/re-creating the default admin account, but it describes a different approach: https://github.com/webanno/webanno/issues/70

Mind that development on WebAnno has largely ceased and efforts have been focussing on [INCEpTION](https://inception-project.github.io) instead. Most WebAnno projects can be simply exported from WebAnno and then imported into INCEpTION.

The issue of resetting/recreating the default admin account has also been addressed already in INCEpTION: https://github.com/inception-project/inception/issues/2503

Also, INCEpTION has a way of setting a default admin PW for unsupervised deployments: https://github.com/inception-project/inception/issues/2164

reckart:
@munterkalmsteiner ?

munterkalmsteiner:

PR #1918, opened 2021-11-27
Nested code blocks should not be used

mohdalani:
we can use Nested code blocks for creating new scope or maybe for restricting the variables defined visibility inside. So here we do not need to use it


munterkalmsteiner:
@mohdalani: the commit contains changes that are irrelevant for the refactoring you propose. Unable to review.

PR #1919, opened 2021-11-27
Methods should not call same-class methods with incompatible "@Transa…

mohdalani:
It is incompatible @Transactional requirement here and if we will use, we will get an exceptions due to Spring  can see only caller and can not do provisions for Calle invoking.


reckart:
I have never seen any exception. What do you mean by "incompatible"?

munterkalmsteiner:
@mohdalani The commit contains changes that are irrelevant for the refactoring you propose. Furthermore, this is not a refactoring as it actually changes the behavior of the code. I don't understand why the annotation is not correct.

PR #1920, opened 2021-11-27
Math operands should be cast before assignment

mohdalani:
It is simple arithmetic operation, so we do not need it because if multiplication result is assigned here for long, it may have already overflowed before the assignment. Instead we can promoted a final type before performed the operation

### DIFF
--- a/webanno-api-dao/src/main/java/de/tudarmstadt/ukp/clarin/webanno/api/dao/AnnotationSchemaServiceImpl.java
+++ b/webanno-api-dao/src/main/java/de/tudarmstadt/ukp/clarin/webanno/api/dao/AnnotationSchemaServiceImpl.java
@@ -2,13 +2,13 @@
  * Licensed to the Technische Universität Darmstadt under one
  * or more contributor license agreements.  See the NOTICE file
  * distributed with this work for additional information
- * regarding copyright ownership.  The Technische Universität Darmstadt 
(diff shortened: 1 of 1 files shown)


webanno-api-dao/src/main/java/de/tudarmstadt/ukp/clarin/webanno/api/dao/AnnotationSchemaServiceImpl.java
```diff
@@ -130,15 +130,15 @@
...
+                                       FeatureSupportRegistry aFeatureSupportRegistry,
+                                       ApplicationEventPublisher aApplicationEventPublisher)
     {
         layerSupportRegistry = aLayerSupportRegistry;
         featureSupportRegistry = aFeatureSupportRegistry;
         applicationEventPublisher = aApplicationEventPublisher;
 
         immutableTagsCache = Caffeine.newBuilder().expireAfterAccess(5, MINUTES)
-                .maximumSize(10 * 1024).build(this::loadImmutableTags);
+                .maximumSize(10240).build(this::loadImmutableTags);
```
reckart, line 141: The compiler most likely converts this number. I expect that the final compiled class number is 10240. The notation is used here for readability.

munterkalmsteiner:
@mohdalani See reckart's comment. The notation increases readability of the code, hence the refactoring is counter-productive in this case.  Contains also unrelated changes.

PR #1921, opened 2021-11-27
Labels should not be used

mohdalani:
Usually the labels are not used in java commonly as well as usually the developers  are not able to understand it. On the other hand, their usage will reduces the code’s readability because it makes the control flow harder to follow


munterkalmsteiner:
@mohdalani This changes the actual behavior of the code, i.e. is not a refactoring. Contains also unrelated changes.

PR #1922, opened 2021-11-27

mohdalani:


munterkalmsteiner:
@mohdalani In principle yes, but why not do this for all parameter literals ("name" is missing). Contains also unrelated changes.

PR #1923, opened 2021-11-28
Track uses of "FIXME" tags

mohdalani:
It is not necessary to use FIXME tags because usually the developer will forget to get back to that tag.


munterkalmsteiner:
@mohdalani In principle correct, but this depends a lot on the project culture. The correct change here would be to create a issue report first, and then delete the tag. Contains also unrelated changes.

PR #1924, opened 2021-11-28
Printf-style format strings should be used correctly

mohdalani:
We must use the format of the Printf-style strings correctly because interpreted at runtime instead of validated by the compiler. It can include errors that occur in the wrong strings.


munterkalmsteiner:
@mohdalani I can't find the refactoring you are referring to due to all the unrelated changes in the commit.

PR #1925, opened 2021-11-28
URIs should not be hardcoded

mohdalani:
Using the URL here will make a Hard code and it will be hard for testing the software. Rather than, we can replaced by customizable parameter.


munterkalmsteiner:
@mohdalani This changes the behavior of the code, i.e. is not a refactoring.

PR #1926, opened 2021-11-28
Non-public methods should not be "@Transactional"

mohdalani:
If we need to use @Transactional the  method must be non-public. Rather than, it will be useless and misleading due to the Spring doesn’t see non-public methods. In addition, it makes no provision for their proper invocation


munterkalmsteiner:
@mohdalani In principle, yes, but the correct solution would be to remove the transactional annotation since this method should be private (at least what I understand from reading the code).

PR #1927, opened 2021-11-28

mohdalani:


munterkalmsteiner:
@mohdalani In principle, yes. But it is not clear why certain literals have been refactored, and some not. In particular, some SQL fragments are not handled in variables, some are still literals. IMO, the lack of consistency here is worse than using literals, for the sake of maintainability.

PR #1928, opened 2021-11-28
Sections of code should not be commented out

mohdalani:
Developers must delete unused code because it leads to software being huge as well as reduces readability.


munterkalmsteiner:
@mohdalani Why not deleting the corresponding comment on the code too? Now the comment does not make sense anymore.

PR #1931, opened 2021-11-29
Annotation Detail pannel placement

lokeshmekala:
Changing the annotation detail placement
The annotation bar is placed in the right corner, and all annotations are present directly above the words. The annotating process would be speed up if the bar was closer to the hover. We can access the annotation easier and fast.
Remove the annotation panel from the right corner of the annotation screen and make the pannel visible whenever the mouse is moved over the annotation. The pannel can be made visible near the annotation. However, the annotation pannel updates only when a double-clicked operation is performed on the annotation. If the mouse is dragged to perform a selection for annotating, the pannel will not be visible to the user, The pannel gets visible when the new annotation is hovered. When the document is opened for the first time, the pannel is empty hence the double click operation must be performed to update the pannel.

### DIFF
--- a/webanno-brat/src/main/java/de/tudarmstadt/ukp/clarin/webanno/brat/resource/visualizer_ui.js
+++ b/webanno-brat/src/main/java/de/tudarmstadt/ukp/clarin/webanno/brat/resource/visualizer_ui.js
@@ -292,6 +292,19 @@ var VisualizerUI = (function($, window, undefined) {
                     if (y < 0) y = 0;
                     if (x < 0) x = 0;
         			element.css({ top: y, left: x });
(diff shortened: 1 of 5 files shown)


reckart:
@lokeshmekala Thank you for the PR. Please note that this repo has essentially reached the end of its life and the development continues in https://inception-project.github.io. If you are still using WebAnno, please try out INCEpTION and consider cross-grading. If there is a reason why you cannot cross-grade, please let me know. Note that many of the issues that are still open in this repo also apply to INCEpTION.

reckart:
@munterkalmsteiner ?

munterkalmsteiner:

PR #1953, opened 2022-01-11
Update DocumentServiceImpl.java

mohdalani:
URIs should not be hardcoded

Because the hard-coded path delimiter "/" makes the code harder to maintain, it must be changed. It also makes cross-platform interoperability more difficult. Hard coding a URI makes it harder to test a program since path literals are not always transferrable between operating systems. for solving this problem i used variable File.separator


reckart:
@munterkalmsteiner ?

munterkalmsteiner:
@mohdalani PR is ok, but you introduce unnecessary formatting changes.

PR #1954, opened 2022-01-11
Update AnnotationSchemaServiceImpl.java

mohdalani:


reckart:
@munterkalmsteiner ?

munterkalmsteiner:
@mohdalani Your PR mixes again different concerns. The one you describe in this PR is ok though.

